Issue #2, opened 2017-11-27
Improve safety of pin API


hannobraun:
The current pin API doesn't do much to prevent the user from making mistakes. It should statically guarantee that it is used correctly.

Off the top of my head, I can think of the following improvements:
- Prevent the user from creating multiple instances of a pin
- Encode pin mode as a type parameter (fixed function, movable function, GPIO, ...)
- Modify peripheral APIs to only work with pins they control

PR #31, opened 2018-01-16
Update pin API

hannobraun:
This pull request updates the pin API, making it safer to use by adding type state for GPIO, as well as laying groundwork for additional SWM-related safety improvements later.

This is another step towards fixing #2, but we're not quite there yet.


hannobraun:
Finally found the time to review this.

The Travis build is currently broken due to unrelated changes to upstream dependencies. I'm going to merge anyway, as `master` is already broken and I can confirm this pull request builds with the old dependency versions.

I'll update the dependencies soon in a separate pull request.

PR #70, opened 2018-05-03
Simplify implementations of SYSCON traits

hannobraun:


hannobraun:
The Travis build fails due to changes to the latest nightly. I'm still merging, as it builds locally, and opened #71 to track the problem.

Issue #76, opened 2018-05-04
Redesign SWM API


hannobraun:
I'm unhappy with some aspects of the `swm` module's design:
- [x] The distribution of responsibility between hand-written and generated code: #48
- [x] The number of traits for fixed and movable functions. Once #48 has been implemented, I believe we can reduce this to one trait for fixed and one trait for movable functions.
- [x] The relationship of `AdcChannel`/`InputFunction`/`OutputFunction` to the traits for fixed and movable functions. There are so many traits, and a lot of complexity when using them in function signatures. I don't have a clear vision here, yet, only the vague feeling that there must be a better way.
- [x] The public API for the SWM lives in the `gpio` module, which is weird at best, confusing at worst. Once #48 is implemented, it can be moved to the new `FixedFunction`/`MovableFunction` types.
- [ ] The `gpio` and `swm` modules are entangled with each other. I think it would be less complex, if `swm` owned `Pin` and related types, and `gpio` was restricted to GPIO APIs.

hannobraun:
Updated the original comment. Some items have been completed.

hannobraun:
Updated original comment. Thanks to #90 and #91, the API has been radically simplified. There are only 2 methods total now (down from 8): One for assigning functions, one for unassigning them.

hannobraun:
Updated original comment. #92 moved the function assignment API to the `swm` module. I also added another item about the entanglement between the `gpio` and `swm` modules.

PR #92, opened 2018-05-22
Move function (un)assignment API from `Pin` to `Function`

hannobraun:
This simplifies the implementation and moves this SWM-related API to the module where it logically belongs.


hannobraun:
Forgot to mention: This is another step towards solving #76.

PR #142, opened 2019-03-12
Update code to conform with latest library changes

mre:
Had to make some changes to get the example to compile. I assume that's because of some recent library changes, but I might be doing it wrong.


mre:
Actually, screw that. I was doing it wrong. 😆 
Was using version 0.2 of the crate.
We should perhaps update the docs for version 0.3 however as it still points to 0.2:
https://docs.rs/lpc82x-hal/0.3.0/lpc82x_hal/#runtime-support

PR #143, opened 2019-03-12
Use an existing name for binary in example call

mre:


hannobraun:
Thank you for the pull request, @mre, but actually this script is obsolete. With the OpenOCD and GDB configuration files in this repository, it's possible now to flash programs using `cargo run`.

I've opened #144 to remove the script.

Issue #137, opened 2019-03-06
LPC845 support


david-sawatzke:
The lpc845 has a pretty nifty little [dev board](https://www.arrow.com/en/products/lpc845-brk/nxp-semiconductors), and it seems pretty similar to the lpc82x. The remaining peripheral differences seem pretty easy to feature-gate, but unfortunately lots of fields are differently named in the `lpc845.svd` compared to the old `lpc82x.svd`, so it's not that easy. 

Fortunately the new `lpc82x.svd` seems more similar, even though there are still some differences left.

A fork seems like a bad idea and I think if we use the new `lpc82x.svd` file, the feature gates aren't going to be too bad. Since the `lpc845.svd` file has to be fixed anyway (since i there doesn't seem to exist an old-variant file) and the new `lpc82x.svd` file is pretty similar, that's seems like the way to go.

hannobraun:
I'd love to extend the `lpx82x-hal` code base to support more devices from the LPC800 family!

Before I address your comments specifically, I'd like to lay out a plan for how that could be done. I think the main problem in supporting a cross-device HAL is re-using as much code as possible, without making the development or usage for any specific device too complicated. The best approach that I know to achieve that is the one used in [`nrf52-hal`](https://github.com/nrf-rs/nrf52-hal). Basically, it works like this:

- There is one crate for common code. In our case this would be called `lpc8xx-hal-common`, or `lpc800-common`. The purpose of this crate is to be a building block for more specific crates, not to be used by end users.
- There are additional crates for each device: `lpc82x-hal` and `lpc84x-hal`. Those depend on `lpc8xx-hal-common` and re-export common code (via `pub use`) in their own API.

I think this approach is very attractive because it makes it easy to share code, while at the same code making it very easy to write device-specific code, where that's necessary. For example, if someone contributes a peripheral API, but can only test on LPC84x, it can be submitted to `lpc84x-hal`, and ported to `lpc8xx-hal-common` later.

Let me know what you think of this approach. I think we have made good experiences with it over at [`nrf52-hal`](https://github.com/nrf-rs/nrf52-hal). The main problem is probably that there's some duplication between the crates (Cargo.toml and such), and that there are multiple crates to build and publish.

---

[...]

PR #150, opened 2019-03-23
Move to nrf52-hal repo layout (#137)

david-sawatzke:
- [x] Introduce a `lpc8xx-hal-common` crate
- [x] Add readme, travis script,etc
- [x] Add lpc845-hal crate

(From https://github.com/lpc-rs/lpc82x-hal/issues/137#issuecomment-470921211)


david-sawatzke:
Is it ok if i just remove contributing.md?

We also need to figure out what to do about changelogs. I think maintaining multiple changelogs (as you suggested in https://github.com/nrf-rs/nrf52-hal/issues/59) is annoying, I'd prefer a global one and keeping the version numbers in sync.

david-sawatzke:
The build script & thus travis seems to work just fine with workspaces without any modifications. I think the only reason the nrf52-hal travis config doesn't use this is because the hals have different cpu architectures.

lpc845-hal/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+documentation = "https://docs.rs/lpc845-hal"
+repository    = "https://github.com/lpc-rs/lpc82x-hal"
+license       = "0BSD"
+readme        = "README.md"
+categories    = ["embedded", "hardware-support", "no-std"]
+keywords      = ["arm", "cortex-m", "nxp", "lpc", "embedded-hal-impl"]
+
+
+[badges]
+maintenance = { status = "actively-developed" }
```
hannobraun: I've been using these badges for a while, but changed my opinion since and started removing them from my projects. Two reasons for this:
1. I think they're too coarse-grained. Is a project that I update once a year actively developed? Not really, but it doesn't fit any of the other categories either.
2. If a project becomes unmaintained, I think it's likely that those badges become just as unmaintained as the project (i.e. the former maintainer forgets to update them).

I don't feel strongly about this, so feel free to keep it, if you want to. I'd prefer to remove it though.

lpc845-hal/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name    = "lpc845-hal"
+version = "0.4.0"
+authors = [
+    "David Sawatzke <david@sawatzke.de>",
+    "Hanno Braun <hanno@braun-robotics.com>"
+]
+edition = "2018"
+
+description   = "Hardware Abstraction Layer (HAL) for the NXP LPC82x series of ARM Cortex-M0+ microcontroller"
```
hannobraun:
```suggestion
description   = "Hardware Abstraction Layer (HAL) for the NXP LPC845 ARM Cortex-M0+ microcontroller"
```

lpc845-hal/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+lpc845-pac   = "0.1"
+nb           = "0.1.1"
+void         = { version = "1", default-features = false }
+# This should be in [dev-dependencies], but those can't be optional. Issue:
+# https://github.com/rust-lang/cargo/issues/1596
+compiletest_rs = { version = "0.3", optional = true }
+
+[dependencies.lpc8xx-hal-common]
+path = "../lpc8xx-hal-common"
+features = ["845"]
```
hannobraun: I'm pretty sure we won't be able to publish it on crates.io like this, due to the `path` dependency. I believe it's possible to combine `path` and `version` though. Then Cargo will know which crates.io dependency to use when downloading `lpc845-hal` as a dependency, but will use the `path` dependency when building locally.

lpc82x-hal/Cargo.toml
```diff
@@ -0,0 +1,70 @@
...
+lpc82x-pac   = "0.6"
+nb           = "0.1.1"
+void         = { version = "1", default-features = false }
+# This should be in [dev-dependencies], but those can't be optional. Issue:
+# https://github.com/rust-lang/cargo/issues/1596
+compiletest_rs = { version = "0.3", optional = true }
+
+[dependencies.lpc8xx-hal-common]
+path = "../lpc8xx-hal-common"
+features = ["82x"]
```
hannobraun: Same comment applies that I left in the other `Cargo.toml`. I believe we need to combine both `path` and `version` here, otherwise we can't publish it on crates.io.

scripts/build.sh
```diff
@@ -10,5 +10,9 @@ TARGET=x86_64-unknown-linux-gnu
 
 # Need to clean to work around this issue:
 # https://github.com/braun-robotics/rust-lpc82x-hal/issues/105
-cargo test --verbose --target=$TARGET &&
-cargo build --verbose --features="rt" --examples --release
+# We can't just compile the workspace, because features are currently additive
+# (see rust-lang/cargo#4463)
+cargo test --manifest-path lpc82x-hal/Cargo.toml --verbose --target=$TARGET &&
+cargo build --manifest-path lpc82x-hal/Cargo.toml --verbose --features="rt" --examples --release
```
hannobraun:
```suggestion
cargo build --manifest-path lpc82x-hal/Cargo.toml --verbose --features="rt" --examples --release &&
```
Otherwise CI will be green, even when the `lpc82x-hal` build fails.

hannobraun:
@david-sawatzke
> Is it ok if i just remove contributing.md?

We should keep it. The commit guidelines are obsolete and should be removed, but the rest is useful. We need to update the release procedure though. That doesn't need to be part of this PR though.

> We also need to figure out what to do about changelogs. I think maintaining multiple changelogs (as you suggested in [nrf-rs/nrf52-hal#59](https://github.com/nrf-rs/nrf52-hal/issues/59)) is annoying, I'd prefer a global one and keeping the version numbers in sync.

I like that!

scripts/build.sh
david-sawatzke: Used `set -e` now, seems better to me in this case

hannobraun: :+1:

lpc845-hal/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name    = "lpc845-hal"
+version = "0.4.0"
+authors = [
+    "David Sawatzke <david@sawatzke.de>",
+    "Hanno Braun <hanno@braun-robotics.com>"
+]
+edition = "2018"
+
+description   = "Hardware Abstraction Layer (HAL) for the NXP LPC825 ARM Cortex-M0+ microcontroller"
```
hannobraun:
```suggestion
description   = "Hardware Abstraction Layer (HAL) for the NXP LPC845 ARM Cortex-M0+ microcontroller"
```

hannobraun: Sorry, if this was introduced by my own suggestions :D

hannobraun:
Build is green. Merging!

PR #151, opened 2019-03-25
Move gpio, syscon & swm peripherals to common-hal crate

david-sawatzke:
Move clock, gpio, swm & syscon to common crate
  
This resulted in two major concessions
 - `RegProxy` had to be duplicated, as it's implemented on types from the pac
 - The `new` functions have to be public. I've made the the `GPIO<Enabled>::new` one unsafe, since it could be used to create a non-enable enable gpio peripheral.


lpc82x-hal/src/lib.rs
```diff
@@ -472,7 +448,7 @@ impl Peripherals {
         Peripherals {
             // HAL peripherals
             DMA   : DMA::new(p.DMA),
-            GPIO  : GPIO::new(p.GPIO_PORT),
+            GPIO  : unsafe { GPIO::new(p.GPIO_PORT) },
```
hannobraun: Would be nice to add a comment (maybe above `Peripherals {`) to explain that any usage of unsafe constructors here is safe, as the hardware is in its initial state, which is what the constructor assumes.

lpc8xx-hal-common/src/gpio.rs
```diff
@@ -60,14 +50,14 @@ use crate::{
...
 pub struct GPIO<State = init_state::Enabled> {
-    pub(crate) gpio  : raw::GPIO_PORT,
-               _state: State,
+    pub(crate) gpio: raw::GPIO_PORT,
+    _state: State,
 }
 
 impl GPIO<init_state::Enabled> {
-    pub(crate) fn new(gpio: raw::GPIO_PORT) -> Self {
+    pub unsafe fn new(gpio: raw::GPIO_PORT) -> Self {
```
hannobraun: Now that the constructor is public, it would be nice to have a bit of documentation. Why (not) to use it, why it's `unsafe`.

lpc8xx-hal-common/src/lib.rs
```diff
@@ -1,7 +1,40 @@
 #![no_std]
 
+use embedded_hal as hal;
 #[cfg(feature = "82x")]
-pub use lpc82x_pac as target;
+pub use lpc82x_pac as raw;
```
hannobraun: I actually prefer `target` to `raw`, but don't feel strongly about it. Do as you think is best!

hannobraun:
> * `RegProxy` had to be duplicated, as it's implemented on types from the pac

I'm not sure I understand the problem. Can't `RegProxy` just be made public in `lpc8xx-hal-common`?

> * The `new` functions have to be public. I've made the the `GPIO<Enabled>::new` one unsafe, since it could be used to create a non-enable enable gpio peripheral.

Find by me. I think that's the most reasonable solution.

Aside from the Travis failure and the `RegProxy` issue, I'd be happy to merge what you have so far and do the rest of your list on follow-up PRs.

lpc8xx-hal-common/src/lib.rs
david-sawatzke: I'll keep raw for now, so that the name is consistent between both crates

lpc8xx-hal-common/src/gpio.rs
david-sawatzke: Added a note explaining that an already enabled gpio port is assumed

david-sawatzke:
> I'm not sure I understand the problem. Can't RegProxy just be made public in lpc8xx-hal-common?

Unfortunately not, we're implementing `RegProxy` on types from `raw`. So it can only be done if `RegProxy` is declared in the same crate. 

It *should* compile now

hannobraun:
> > I'm not sure I understand the problem. Can't RegProxy just be made public in lpc8xx-hal-common?
> 
> Unfortunately not, we're implementing `RegProxy` on types from `raw`. So it can only be done if `RegProxy` is declared in the same crate.
> 
> It _should_ compile now

Ah, got it. I need to look into this at some point. There's probably a better solution.

PR #153, opened 2019-03-29
Remove maintanance badge

hannobraun:
Based on #152, so shouldn't be merged before that one has been merged.


hannobraun:
Merging without review, since this only concerns `lpc82x-hal`.

PR #154, opened 2019-03-29
WIP: Build failure experiments

david-sawatzke:
The workspace setup seems to have introduced a broken link setup. But since the ci script overwrote the rustflags and thus didn't use the proper link file, I didn't catch it.
The issue seems that with the workspace setup it only searches the linkerfile in the crate root.
Solutions:
- Add the `memory.x` to the workspace root, so it gets picked up
- Add a build file to each hal to copy it to the build location
- Move away from the workspace setup (I personally prefer this one, it doesn't seem to add a lot of benefits while adding additional headache)

How to fix the CI script:
- Just don't overwrite `rustflags`, leads to warnings slipping through
- Add `-C link-arg=-Tlink.x` to it for the builds

I'm not quite happy with both, so I'm very open to suggestions

(Gonna take a look at the ci failures later on, works locally)


hannobraun:
Sorry for the delay.

> The workspace setup seems to have introduced a broken link setup. But since the ci script overwrote the rustflags and thus didn't use the proper link file, I didn't catch it.

Can you what's happening here? I don't understand how settings `RUSTFLAGS` affects which linker script is used. Maybe some weird behavior in Cargo?


> The issue seems that with the workspace setup it only searches the linkerfile in the crate root.
> Solutions:
> 
> * Add the `memory.x` to the workspace root, so it gets picked up
> 
> * Add a build file to each hal to copy it to the build location
> 
> * Move away from the workspace setup (I personally prefer this one, it doesn't seem to add a lot of benefits while adding additional headache)

I agree that removing the workspace is the best solution.

> How to fix the CI script:
> 
> * Just don't overwrite `rustflags`, leads to warnings slipping through
> 
> * Add `-C link-arg=-Tlink.x` to it for the builds
> 
> 
> I'm not quite happy with both, so I'm very open to suggestions

I also don't like either. Don't have any suggestions right now.

david-sawatzke:
> I don't understand how settings RUSTFLAGS affects which linker script is used.

In `.cargo/config` rust is told which link script to use via `rustflags`, which is (probably) overwritten
```
[target.thumbv6m-none-eabi]
runner = "arm-none-eabi-gdb"
rustflags = [
    "-C", "link-arg=-Tlink.x",
]
```

hannobraun:
Ah, I get it now. The `RUSTFLAGS` environment variable and `rustflags` in `.cargo/config` are just two ways to do the same thing, but `RUSTFLAGS` [takes precedent](https://doc.rust-lang.org/cargo/reference/config.html#environment-variables). I had assumed that both would be merged.

I guess the best option for now is to remove `RUSTFLAGS` from the build script. I don't really like that, but duplicating the linker script configuration will lead to problems, if that configuration ever needs to change.

We could re-add `#[deny(warnings)]` to the code. I previously removed it [according to this guideline](https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md). I remember having problems with `#[deny(warnings)]` in the past, but to be honest, I don't remember which problems specifically. Maybe we can just add it and see how it works out.

david-sawatzke:
> I don't really like that, but duplicating the linker script configuration will lead to problems, if that configuration ever needs to change.

It may lead to problems, but most of them will result in the failure of the ci script and are thus easily fixed.

I think adding `#[deny(warnings)]` is worse than that, for the reasons given in the guideline and also encouraging temporary (and thus eventually permanent) removal for development, since warnings often occur for e.g. unused functions.

hannobraun:
> > I don't really like that, but duplicating the linker script configuration will lead to problems, if that configuration ever needs to change.
> 
> It may lead to problems (although likely not in the near future), but most of them will result in the failure of the ci script and are thus easily fixed. We could add a warning in `.cargo/config`.
> 
> I think adding `#[deny(warnings)]` is worse than that, for the reasons given in the guideline and also encouraging temporary (and thus eventually permanent) removal for development, since warnings often occur for e.g. unused functions.

Both very good points!

PR #156, opened 2019-04-05
Clean up leftovers from move to multi-crate layout

david-sawatzke:
This also
- Removes the workspace setup
- Fixes the ci script, by adding the linker script to the RUSTFLAGS

Resolves #154


hannobraun:
Looks good to me. Thank you, @david-sawatzke!

PR #157, opened 2019-04-18
Re-add workspace

hannobraun:


hannobraun:
Damn, so the workspace messes up the build just by being there. I thought we'd be fine if we kept using the build script for building.

Sorry for the noise.

PR #158, opened 2019-05-29
Update reference to Braun Robotics

hannobraun:
@lpc-rs/all Please take a look at https://github.com/lpc-rs/lpc-pac/pull/34, which explains the reasoning behind this change and why I want to run it by you.


hannobraun:
Thanks guys!

The Travis failure is unrelated, and it's pretty much impossible that this change is going to break anything. Merging.

PR #161, opened 2019-10-11
Merge all crates into one

hannobraun:
I think that the previous approach to portability hasn't worked out
well. Having three crates was more complicated, led to duplicated code,
and it wasn't even successful: The common crate still contained
conditional code, depending on which target was compiled.

This commit collapses the three crates into a single one that uses
conditional compilation all the way. This is just a first step, but I
believe it's already better than what we had before.


hannobraun:
I'm merging this now. As I said in the other PR, I need a new release out today.

@david-sawatzke I hope it's okay that I'm merging/releasing without a review. I'm happy to discuss and make change later, as required.

PR #163, opened 2019-10-15
LPC845 fixup

david-sawatzke:
This:
- (Re-)adds the lpc845 gpio example
- Add correct openocd config

TODO:
`rt` feature doesn't work. Both pac crates seem to added as dependencies, with the lpc82x-pac `device.x` being used on my machine, regardless of the selected mcu. This results in linker failure. 

The simplest fix appears to be adding two separate `rt` features, but that might result in user confusion

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,10 @@ name              = "gpio"
 required-features = ["rt", "82x"]
 
 [[example]]
+name              = "845_gpio"
+required-features = ["rt", "845"]
+
+[[example]]
 name              = "i2c_vl53l0x"
 required-features = ["rt", "82x"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,8 @@ default = ["82x"]
 # This is needed to make the compiletest stuff optional. It requires std, which
 # means we can't build it together with the examples.
 compiletest = ["compiletest_rs"]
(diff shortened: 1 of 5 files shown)


hannobraun:
Thanks for the PR!

> TODO:
> `rt` feature doesn't work. Both pac crates seem to added as dependencies, with the lpc82x-pac `device.x` being used on my machine, regardless of the selected mcu. This results in linker failure.
> 
> The simplest fix appears to be adding two separate `rt` features, but that might result in user confusion

That's very unfortunate. This problem doesn't show up in the stm32-rs world, as they bundle multiple PACs into one crate (e.g. `stm32l0` for STM32L0x1/STM32L0x2/STM32L0x3).

I'll have to think about this some more. Maybe there's a good solution within this crate. If not, we might want to create a single `lpc8xx-pac` crate, like stm32-rs does.

david-sawatzke:
@hannobraun I'd like to merge this now (with the "temporary" `rt_845` option), would that be okay for you?

hannobraun:
Yes, I think that would be best. I've experimented with this too, and I think short of merging the PACs, separate `rt` features is the best we can do.

I have some nitpicks about this pull request, but I have some in-progress work anyway, so I'll just merge this now and submit my improvements as a separate PR.

Thanks!

PR #165, opened 2019-10-22
Improve LPC845 support

hannobraun:
@david-sawatzke This is based on your previous pull request and my own experimentation. I think it's much nicer than before. Please let me know what you think!

I have one problem though: When I try to run the GPIO example on LPC845 in release mode, I only get to line 50. If I step one step further, it tries to jump to `0xfffffffe`, which of course doesn't work.

It works in release mode. Have you seen this before?


david-sawatzke:
I've had this issue just now, pressing reset or `mon reset` allowed execution to continue. No idea why.

hannobraun:
Hmm, can confirm. Pressing reset fixes it.

So at least we can run programs in both debug and release modes, but this means that debugging is effectively broken on LPC845.

david-sawatzke:
With `mon reset halt` i seem to land in the bootloader (probably), but debugging works just fine

hannobraun:
Thanks, I accidentally ignored the `mon reset` bit when first reading your comment. I got it to work now!

The only problem I have now is that it requires a Git version of OpenOCD. At least I wasn't able to get the LPC84x configuration from OpenOCD's Git running with the latest released version. Not a problem for myself, but I'm running a workshop at RustFest, and I don't think I can expect people to compile it from source.

Oh well, I'll figure something out.

PR #166, opened 2019-10-22
Add delay implementation based on systick

david-sawatzke:
This is based on https://github.com/stm32-rs/stm32f0xx-hal/blob/master/src/delay.rs, (mostly) written by myself and adjusted to better fit the api.

Also adds a gpio example based on this that's less hacky for the lpc845.

I didn't find a way to get the current system frequency & (for example) i2c seems to rely on it being 12 MHz, so it's just hardcoded for now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,6 +80,10 @@ name              = "gpio_simple"
 required-features = ["rt-selected"]
 
 [[example]]
+name              = "gpio_delay"
+required-features = ["rt-selected"]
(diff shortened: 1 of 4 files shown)


hannobraun:
Thank you!

I'm not a big fan of the extension trait pattern (`SystDelay`) and would instead have just added a constructor to `Delay`. The rest looks good though, and we can improve this later if we wish, so I'm merging it.

PR #167, opened 2019-10-22
Add debug information to release builds & more optimization

david-sawatzke:
Otherwise debugging in release mode doesn't provide the corresponding source code in gdb.

Also optimizes more, so that the release build is as small as possible


hannobraun:
Looking good, thank you!

PR #169, opened 2019-10-23
Simplify `Delay` construction

hannobraun:
@david-sawatzke This is my take on your delay implementation. Please let me know what you think!


david-sawatzke:
I tried to keep a similar style to the rest of the api, but I agree that doing it via extension traits is a bit iffy. I'm also not completely happy with this, but since it's more straightforward it's probably less confusing than trying to emulate the existing api in a different way.

PR #171, opened 2019-10-23
Enable WKT for lpc845

david-sawatzke:
FroDerivedClock is the only thing different, but it's the same as IrcDerivedClock with a different name (no idea why nxp did that).

### DIFF
--- a/examples/gpio_sleep.rs
+++ b/examples/gpio_sleep.rs
@@ -27,16 +27,27 @@ fn main() -> ! {
     let     swm    = p.SWM.split();
     let mut syscon = p.SYSCON.split();
     let mut wkt    = p.WKT.enable(&mut syscon.handle);
+    #[cfg(feature = "82x")]
+    let     gpio   = p.GPIO; // GPIO is initialized by default on LPC82x.
+    #[cfg(feature = "845")]
+    let     gpio   = p.GPIO.enable(&mut syscon.handle);
 
     // We're going to need a clock for sleeping. Let's use the IRC-derived clock
     // that runs at 750 kHz.
+    #[cfg(feature = "82x")]
     let clock = syscon.irc_derived_clock;
+    #[cfg(feature = "845")]
+    let clock = syscon.fro_derived_clock;
 
-    // Configure the PIO0_12 pin. The API tracks the state of pins at
-    // compile-time, to prevent any mistakes.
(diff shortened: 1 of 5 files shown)


examples/gpio_sleep.rs
```diff
@@ -27,16 +27,27 @@ fn main() -> ! {
     let     swm    = p.SWM.split();
     let mut syscon = p.SYSCON.split();
     let mut wkt    = p.WKT.enable(&mut syscon.handle);
+    #[cfg(feature = "82x")]
+    let     gpio   = p.GPIO; // GPIO is initialized by default on LPC82x.
```
hannobraun: I wonder if it's a good idea to disable GPIO on LPC82x in `Peripherals::take`. That's a bit of extra code and extra time for most applications, but it would make applications more portable. I don't know, and it has nothing to do with this pull request anyway. Just something that came to mind.

examples/gpio_sleep.rs
```diff
@@ -27,16 +27,27 @@ fn main() -> ! {
...
+    #[cfg(feature = "845")]
+    let     gpio   = p.GPIO.enable(&mut syscon.handle);
 
     // We're going to need a clock for sleeping. Let's use the IRC-derived clock
     // that runs at 750 kHz.
+    #[cfg(feature = "82x")]
     let clock = syscon.irc_derived_clock;
+    #[cfg(feature = "845")]
+    let clock = syscon.fro_derived_clock;
```
hannobraun: If the clocks are the same, it might be a good idea to normalize the name. Maybe choose one of them, or just choose a new name that's descriptive.

src/syscon.rs
```diff
@@ -70,8 +70,14 @@ impl SYSCON {
...
             bod: BOD(PhantomData),
             flash: FLASH(PhantomData),
+            #[cfg(feature = "82x")]
             irc: IRC(PhantomData),
+            #[cfg(feature = "82x")]
             ircout: IRCOUT(PhantomData),
+            #[cfg(feature = "845")]
+            fro: FRO(PhantomData),
+            #[cfg(feature = "845")]
+            froout: FROOUT(PhantomData),
```
hannobraun: I think it would be better if the code here were the same for both platforms. That would require us to settle on one name (see other comment) and hide the difference in a type definition.

PR #172, opened 2019-10-24
Merge Irc & FRO to an IOSC (internal oscillator)

david-sawatzke:
For our purposes, they're the same.

https://github.com/lpc-rs/lpc8xx-hal/pull/171#discussion_r338108537


david-sawatzke:
I understand your point, but it's still a hardware resource (albeit with a different name), so it should still be named like one (in my opinion).

Since the "true" name is documented mostly everywhere, I think it's not that confusing (since everyone reads the docs /s).

PR #175, opened 2019-11-06
[WIP] Usart implementation

david-sawatzke:
TODO:
- Improve Documentation
- Investigate swm & maybe force the user to enable it on 845

### DIFF
--- a/examples/usart.rs
+++ b/examples/usart.rs
@@ -7,6 +7,9 @@ use lpc8xx_hal::{
     cortex_m_rt::entry, prelude::*, syscon::clocksource::PeripheralClockConfig, Peripherals,
 };
 
+#[cfg(feature = "845")]
+use lpc8xx_hal::pac::syscon::frg::frgclksel::SEL_A;
+
 #[entry]
 fn main() -> ! {
     let p = Peripherals::take().unwrap();
@@ -14,6 +17,8 @@ fn main() -> ! {
     let mut swm = p.SWM.split();
     let mut syscon = p.SYSCON.split();
 
+    // TODO
+    //
     // For some reason, the clock for swm need to be enabled, even though
     // it should be enabled from the start
     swm.handle = swm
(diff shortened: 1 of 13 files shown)


hannobraun:
Oh man, I've been doing pretty much the same :-)

I need to get this done in time for RustFest (need it for the workshop). I might use or ignore your stuff, whatever turns out more convenient. Once I'm back, I'll take a closer look and properly merge both of our work.

david-sawatzke:
@hannobraun The example works now, there were 2 issues:
- The pinout is misleading, 25 is the tx pin
- For some reason, the clock for swm needs to be enabled, even though it should be enable from the start

src/usart.rs
```diff
@@ -447,6 +447,27 @@ impl Peripheral for pac::USART2 {
...
+    const INTERRUPT: Interrupt = Interrupt::PIN_INT6_USART3;
+
+    type Rx = swm::U3_RXD;
+    type Tx = swm::U3_TXD;
+}
+
+#[cfg(feature = "845")]
+impl Peripheral for pac::USART4 {
+    /// USART4 and PIN_INT7 share an interrupt, this may cause difficulties
+    /// when trying to use both at the same time
```
hannobraun: Same of above, might be worth putting that into the documentation of the respective field in `Peripherals`.

examples/usart.rs
```diff
@@ -1,25 +1,26 @@
...
-
 #[entry]
 fn main() -> ! {
     let p = Peripherals::take().unwrap();
 
-    let mut swm    = p.SWM.split();
+    let mut swm = p.SWM.split();
     let mut syscon = p.SYSCON.split();
 
+    // TODO
```
hannobraun:
```suggestion
    // For some reason, the clock for swm needs to be enabled, even though it
    // should be enable from the start.
```

(this is the text from your comment)

examples/usart.rs
```diff
@@ -1,25 +1,26 @@
...
+    swm.handle = swm
+        .handle
+        .disable(&mut syscon.handle)
+        .enable(&mut syscon.handle);
```
hannobraun: I'm fine with merging it like this, given the comment is updated (suggestion above) and we open a separate issue to track this problem.

examples/usart.rs
```diff
@@ -44,25 +51,71 @@ fn main() -> ! {
...
+    #[cfg(feature = "82x")]
+    let rx_pin = swm.pins.pio0_18.into_swm_pin();
+
+    #[cfg(feature = "845")]
+    // Make PIO0_25 and PIO0_24 available to the switch matrix API, by changing
+    // their state using `into_swm_pin`. This is required, because we're going
+    // to use the switch matrix to assigne the USART0 functions to those pins.
+    //
+    // WARNING: The pinout for the lpc845brk uses tx/rx as seen from the
+    // perspective from the serial adapter, so this is used the opposite way
```
hannobraun: Oh man, that's one of these things... probably took ages to figure out :-)

hannobraun:
@david-sawatzke By the way, I'm currently doing some LPC845-related work for a client, which means I have the time to take this over, if you want. I don't want to step on your toes, I'm just offering to get this over the finish line, in case you're busy.

examples/usart.rs
david-sawatzke: Yeah, together with the swm issue, this took quite a while. I'm really thankful the c example code did the right thing

david-sawatzke:
@hannobraun Thanks for the review!

I'm sorry about the code formatting stuff, I probably didn't pay attention when doing that and kept auto formatting on.

The TODOs were replaced by (bare bones) documentation, which should be enough for now.

I'm not happy with the clockconfig stuff, so I'd be happy if you had a go at it (everything else is also fair game).

hannobraun:
Thank you for the additional changes! Merging.

> I'm sorry about the code formatting stuff, I probably didn't pay attention when doing that and kept auto formatting on.

No problem. It's just not ideal, which is why I wanted to mention it.

> I'm not happy with the clockconfig stuff, so I'd be happy if you had a go at it (everything else is also fair game).

I'll take a look. I have some more immediate priorities right now, but I'll take a stab at it later.

PR #184, opened 2020-01-07
Format for rustfmt

david-sawatzke:
TODO:
- Maybe integrate clippy?
  - not sure if that's possible outside github actions
- Nightly & stable styles will (probably) diverge, what then?

As discussed here: https://github.com/lpc-rs/lpc8xx-hal/pull/175#pullrequestreview-338579975

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,4 @@ install:
   - rustup component add rustfmt
 
 script:
-  - |
-    cargo fmt -- --check &&
-    ./scripts/build.sh
+  - ./scripts/build.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ addons:
 
 install:
   - rustup target add thumbv6m-none-eabi
+  - rustup component add rustfmt
 
 script:
-  - ./scripts/build.sh
+  - |
+    cargo fmt -- --check &&
+    ./scripts/build.sh
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -3,6 +3,11 @@ set -e
 
 # Fail build, if there are any warnings.
 export RUSTFLAGS="-D warnings"
+#
+# Check for formatting with the stable rustfmt
+if [ "$TRAVIS_RUST_VERSION" != beta ] &&  [ "$TRAVIS_RUST_VERSION" != nightly ]; then
+    cargo fmt -- --check
+fi
 
 cargo build --verbose --no-default-features --features=82x-rt --examples
 cargo build --verbose --no-default-features --features=845-rt --examples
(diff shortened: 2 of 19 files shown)


.travis.yml
```diff
@@ -12,6 +12,9 @@ addons:
 
 install:
   - rustup target add thumbv6m-none-eabi
+  - rustup component add rustfmt
 
 script:
-  - ./scripts/build.sh
+  - |
+    cargo fmt -- --check &&
+    ./scripts/build.sh
```
hannobraun: Might be better to add this check to `build.sh`, to prevent surprises when the pull request is submitted. Not sure. What do you think?

hannobraun:
>     * Maybe integrate clippy?
>       
>       * not sure if that's possible outside github actions

I think we should integrate Clippy into the build, but that can be done in a separate pull request. Why wouldn't that be possible with Travis?

If this indeed requires GitHub Actions, I'd be happy to switch over. I've tried it in [another project](https://github.com/kari-lang/kari/blob/master/.github/workflows/rust.yml) and I'm happy with it. Works well and seems faster than Travis, although I didn't do any real measurements.

>     * Nightly & stable styles will (probably) diverge, what then?

Any such diversions should be short-term (until the next stable version is released), and a developer can probably work around it easily. I assume it could become a problem on CI though.

I suggest we just wait and see. If we run into it, we can configure Travis to not fail the whole build if only nightly fails, which seems to be a pretty common thing to do.

.travis.yml
hannobraun: I'm not merging until you had a chance to reply to this comment. Feel free to just merge yourself, once you've addressed this or decided not to.

scripts/build.sh
```diff
@@ -3,6 +3,11 @@ set -e
 
 # Fail build, if there are any warnings.
 export RUSTFLAGS="-D warnings"
+#
+# Check for formatting with the stable rustfmt
+if [ "$TRAVIS_RUST_VERSION" != beta ] &&  [ "$TRAVIS_RUST_VERSION" != nightly ]; then
+    cargo fmt -- --check
+fi
```
hannobraun: Nice solution! :+1:

hannobraun:
And Travis is green. Merging.

david-sawatzke:
> I think we should integrate Clippy into the build, but that can be done in a separate pull request. Why wouldn't that be possible with Travis?

I don't wanna fail the build with clippy warnings, just annotate the PR, like here: https://github.com/probe-rs/probe-rs/pull/96/commits/e60fcf3e0f26548962b0512f86627b74a9c4e614, and I'm not sure if that's possible with travis (based on a cursory look)

hannobraun:
Ah, I see. That's pretty neat!

PR #185, opened 2020-01-08
Fix panic in build script, if target doesn't exist

hannobraun:
I ran into this when starting a new project depending on LPC8xx HAL.
Seems pretty serious, since this might be the first point of contact
with this library for many users.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,7 @@
-use std::{fs::File, io::prelude::*};
+use std::{
+    fs::{self, File},
+    io::prelude::*,
+};
 
 use termion::{color, style};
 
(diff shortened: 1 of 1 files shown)


hannobraun:
Hold on, there's still something wrong. Everything works if I use a path dependency on my fixed version, but I still get the same error when depending on the Git version, even though my target directory exists now. Investigating.

hannobraun:
Okay, the reason is that `build.rs` isn't run from the current directory, but from wherever it is located. The main purpose of writing `openocd.cfg` in the build script is for use by the examples in this repository, so it doesn't really matter that the resulting configuration isn't accessible in the user's `target` directory (although it would have been useful).

I've added a comment to `build.rs`, explaining why `OUT_DIR` isn't used to write to the user's target directory, as that is the question that came up during my investigation. This is ready to review/merge now.

Issue #14, opened 2017-11-28
Make runtime support more convenient


hannobraun:
Currently, the user must specify the MCUs memory layout (by adding `memory.x`). This could be made more convenient, by allowing the user to just specify which part they use (maybe using a Cargo feature?) and somehow creating the memory layout for [cortex-m-rt](https://crates.io/crates/cortex-m-rt) automatically.

hannobraun:
There's some work on this in #187, although that selects a conservative configuration that should fit all target devices. It would be better, if it selected a correct configuration based on Cargo features specified.

david-sawatzke:
I'm not sure if that's a good idea to select a memory.x file based on features, especially since we'd need to add an additional feature differentiating between lpc822 & lpc824.

I also think not "hiding" the need for a memory.x file is the better choice for hals, but that's just personal preference.

hannobraun:
> I'm not sure if that's a good idea to select a memory.x file based on features, especially since we'd need to add an additional feature differentiating between lpc822 & lpc824.

I think having additional features to differentiate between hardware differences is a good thing. That way, the code can be configured to provide exactly what the user needs, and prevent accidental use of features that aren't actually there.

Granted, I haven't looked at the differences between LPC822 and LPC824 lately, so I don't need how relevant this would be besides memory. But I think memory is a good enough reason to add those features, and I don't think they would be too much of a burden (existing hardware doesn't change much, after all).

> I also think not "hiding" the need for a memory.x file is the better choice for hals, but that's just personal preference.

Two notes:
- Right now, a first point of contact between new users and LPC8xx HAL is an obscure error message about `memory.x`. We can fix that problem, and I think we should.
- For most users (i.e. those running a single application that doesn't do anything weird memory-wise), there really is no need for providing their own `memory.x`, if the correct one is selected for their hardware. Forcing users to figure this out themselves is neither necessary nor user-friendly.

Would your concerns be alleviated, if I extended my pull request to provide the correct configuration for all supported targets?

> I think we should first let the discussion play out in matrix.

I don't like this suggestion :-)

First, I don't have the time nor inclination to follow that discussion. I think the format is good for socializing and real-time stuff, but not for drawn-out technical discussions.

Second, as far as I did follow it, the discussion was about shortcomings in STM32L0 HAL. We can simply do a better job here.

[...]

PR #187, opened 2020-01-08
Make further improvements to `build.rs`

hannobraun:
Based on #186. I recommend reviewing/merging that one first.


hannobraun:
Rebased, ready for review.

build.rs
```diff
@@ -25,11 +37,45 @@ fn main() {
     // purpose. `OUT_DIR` points to a directory within target, whose path even
     // contains a hash. This configuration file needs to be referenced from the
     // GDB configuration, which can't just ask Cargo where to look for it.
+    fs::create_dir_all("target")?;
+    File::create("target/openocd.cfg")?.write_all(openocd_cfg)?;
```
david-sawatzke: If the errors are just bubbled up, why not just `.unwrap()`?

hannobraun: I think this reads more nicely. `unwrap` and `expect` breaks the flow of what's actually supposed to happen.

In general, I believe that lower layers should report errors, and leave it to upper layers to handle them. But granted, that doesn't make much of a difference in this tiny script.

hannobraun:
(continuing the discussion from #14)

@david-sawatzke I went ahead and implemented full target selection support. I see that you commented in #14 and approved this pull request in the meantime. I still think this was worthwhile work.

Please let me know what you think about the new commits.

hannobraun:
I just saw that the order of commits, as displayed here, is wrong. The first two commits should be switched. The reason for that is, that I regularly rebase while I develop locally, to make the resulting commits easier to read. Unfortunately, GitHub shows commits in pull requests sorted by time, not by actual order. Please rest assured, that the actual order (as shown by `git log`) is correct.

memory_32_8.x
```diff
@@ -0,0 +1,9 @@
+/*
+    Set memory sizes to the smallest values found in any of the  LPC800 parts we
```
david-sawatzke: This should be updated, since it's a specific value now (same in the 2 other memory.x files)

david-sawatzke:
We also need to featuregate CAPT & DAC1 now (probably also previously, but now more obviously)

memory_32_8.x
hannobraun: Right. I've pushed another commit to remove them.

hannobraun:
Merging now, since the previous commit was approved, and the last one only removes some outdated comments.

> We also need to featuregate CAPT & DAC1 now (probably also previously, but now more obviously)

I've opened #191.

PR #189, opened 2020-01-08
Implement digital input

david-sawatzke:
Also adds an example, currently only for lpc845

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -304,7 +304,6 @@ where
     ///
     /// # Example
     ///
-    /// # TODO
     /// ``` no_run
     /// use lpc82x_hal::prelude::*;
     /// use lpc82x_hal::Peripherals;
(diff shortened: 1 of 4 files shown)


hannobraun:
Thank you! I didn't get to it today, but I intend to take a closer look at this tomorrow.

src/gpio.rs
```diff
@@ -279,7 +281,105 @@ where
...
+    /// This method is only available, if the pin is in the GPIO state and the
+    /// pin is not already in input mode, i.e. the pin direction is output or
+    /// unknown. You can enter the GPIO state using [`Pin::into_gpio_pin`].
+    ///
+    /// Consumes the pin instance and returns a new instance that is in output
+    /// mode, making the methods to set the output level available.
+    ///
+    /// # Example
+    ///
+    /// # TODO
```
hannobraun: Errant `TODO` spotted. Maybe referring to the example, which is out of date?

src/gpio.rs
```diff
@@ -279,7 +281,105 @@ where
...
+    ///
+    /// // Transition pin into GPIO state, then set it to output
+    /// let mut pin = swm.pins.pio0_12
+    ///     .into_gpio_pin(&p.GPIO)
+    ///     .into_output();
+    ///
+    /// // Output level can now be controlled
+    /// pin.set_high();
+    /// pin.set_low();
+    /// ```
```
hannobraun: This example is out of date. This should have been spotted by CI.

I'm fine with merging this as-is, and later fix the examples in one go. I've opened #192 to track this.

src/gpio.rs
```diff
@@ -279,7 +281,105 @@ where
...
+    pub fn into_input(
+        self,
+    ) -> Pin<T, pin_state::Gpio<'gpio, direction::Input>> {
+        // TODO
```
hannobraun: Errant `TODO`.

src/gpio.rs
david-sawatzke: The example was for the gpio output, but it's out of date as well

PR #190, opened 2020-01-09
Force swm to be enabled on lpc845, fix #183

### LINKED ISSUE #183: Investigate SWM weirdness
Even though, according to the reference manual, swm should be enabled per default, it needs to be reenabled in the [usart example](https://github.com/lpc-rs/lpc8xx-hal/blob/8ca1a9ce19eef65904695a24bb6113420578966d/examples/usart.rs#L20) for the lpc845.

It's listed as disabled per default in the svd.

It shouldn't be possible to use it in a disabled state

david-sawatzke:
Also makes the `new` functions public, to stay in line with gpio and
bypass unused function warnings.

Was there a good reason to implement these on `Handle` instead of the whole peripheral?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -206,7 +206,24 @@ pub struct Peripherals {
     pub PMU: PMU,
 
     /// Switch matrix
-    pub SWM: SWM,
+    ///
+    /// By default, the switch matrix is enabled on the LPC82x and disabled on
+    /// the LPC845.
+    ///
+    /// The reference manual for the LPC845 suggests otherwise, but it seems to
+    /// be wrong.
+    #[cfg(feature = "82x")]
+    pub SWM: SWM<init_state::Enabled>,
+
+    /// Switch matrix
+    ///
+    /// By default, the switch matrix is enabled on the LPC82x and disabled on
+    /// the LPC845.
+    ///
+    /// The reference manual for the LPC845 suggests otherwise, but it seems to
+    /// be wrong.
+    #[cfg(feature = "845")]
+    pub SWM: SWM<init_state::Disabled>,
 
     /// System configuration
     pub SYSCON: SYSCON,
@@ -512,7 +529,10 @@ impl Peripherals {
             #[cfg(feature = "82x")]
             I2C0: I2C::new(p.I2C0),
             PMU: PMU::new(p.PMU),
-            SWM: SWM::new(p.SWM0),
+            #[cfg(feature = "82x")]
+            SWM: unsafe { SWM::new_enabled(p.SWM0) },
+            #[cfg(feature = "845")]
+            SWM: unsafe { SWM::new(p.SWM0) },
             SYSCON: SYSCON::new(p.SYSCON),
             USART0: USART::new(p.USART0),
             USART1: USART::new(p.USART1),
--- a/src/swm.rs
+++ b/src/swm.rs
@@ -30,23 +30,59 @@ use self::pin_state::PinState;
 /// [`swm::Parts`]: struct.Parts.html
 /// [`Peripherals`]: ../struct.Peripherals.html
 /// [module documentation]: index.html
-pub struct SWM {
+pub struct SWM<State = init_state::Enabled> {
     swm: pac::SWM0,
+    state: State,
 }
 
-impl SWM {
-    pub(crate) fn new(swm: pac::SWM0) -> Self {
-        SWM { swm }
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -206,7 +206,14 @@ pub struct Peripherals {
...
     /// Switch matrix
-    pub SWM: SWM,
+    ///
+    /// The Switch matrix is enabled by default.
+    #[cfg(feature = "82x")]
+    pub SWM: SWM<init_state::Enabled>,
+
+    /// Switch matrix
+    #[cfg(feature = "845")]
+    pub SWM: SWM<init_state::Disabled>,
```
hannobraun: I think the documentation is not ideal. Only one version will show up in online documentation, and that is bound to confuse users of the other variant.

Suggestion: Explain the whole situation (default-enabled on LPC82x, not on LPC845), copy-paste the documentation to both variants, and try to keep that in sync forever.

Not ideal, but probably good enough. And I can't think of a better solution (code generation would probably be overkill).

src/swm.rs
```diff
@@ -30,13 +30,64 @@ use self::pin_state::PinState;
...
+pub struct SWM<State = init_state::Enabled> {
     swm: pac::SWM0,
+    _state: State,
 }
 
-impl SWM {
-    pub(crate) fn new(swm: pac::SWM0) -> Self {
-        SWM { swm }
+impl SWM<init_state::Disabled> {
+    /// Create an disabled swm peripheral
```
hannobraun:
```suggestion
    /// Create a disabled SWM peripheral
```

src/swm.rs
```diff
@@ -30,13 +30,64 @@ use self::pin_state::PinState;
...
+        syscon.enable_clock(&self.swm);
+
+        SWM {
+            swm: self.swm,
+            _state: init_state::Enabled(()),
+        }
+    }
+}
+impl SWM<init_state::Enabled> {
+    /// Create an enabled swm peripheral
```
hannobraun:
```suggestion
    /// Create an enabled SWM peripheral
```

src/swm.rs
```diff
@@ -100,67 +174,13 @@ pub struct Parts {
 /// PMU.
 ///
 /// [module documentation]: index.html
-pub struct Handle<State = init_state::Enabled> {
+pub struct Handle {
     swm: pac::SWM0,
-    _state: State,
 }
 
-impl Handle<init_state::Enabled> {
```
hannobraun: I think this should stay on `Handle`. The clock is only required for the switch matrix interface. The peripheral itself doesn't need a clock to function. Configuring the switch matrix, then disabling its clock to save power, is completely valid, but no longer possible with your changes.

david-sawatzke:
> The user might have done anything with the PAC struct, invalidating our type state and all assumptions we make based on it (there's a lot more type state in there than just enabled/disabled).

I'm not sure about it, and it's better to get close to this goal than to never try it, but don't we have issues with IOCON?

With the IOCON, one can set pins to Hi-Z and even worse, on the lpc845 that can happen via the FAIM, so we don't have the option to intercept it.

src/swm.rs
david-sawatzke: Ah, knew there had to be a reason for this

hannobraun:
> I'm not sure about it, and it's better to get close to this goal than to never try it, but don't we have issues with IOCON?
> 
> With the IOCON, one can set pins to Hi-Z and even worse, on the lpc845 that can happen via the FAIM, so we don't have the option to intercept it.

Well, LPC8xx HAL isn't complete yet, and it's not like I've studied all of the user manuals in detail. The API guarantees exist on a best-effort basis. There's not an IOCON API so far, so whatever damage you can do there is out of scope for now. It might be good to add a note to that effect to the GPIO documentation.

Regarding the FAIM, I haven't looked at it in detail, but I think we need to update the affected APIs as follows:
- Add an "unknown" type state.
- Provide a safe method that moves from that unknown state to the user's desired state, doing a runtime check to make sure the hardware settings match that desired state. That would basically return an `Option<Pin<ExpectedState>>`, or maybe a `Result`.
- Maybe provide a variant that sets the expected state (would return `Pin<ExpectedState>` directly).
- Provide an unsafe method that does the same, minus the runtime checking (if the user is sure and doesn't want the overhead). Would return `Pin<ExpectedState>`.

Just some rough thoughts for now. Details to be determined.

PR #195, opened 2020-01-10
Re-export clock source enum from `frg` module

hannobraun:
This makes the API much friendlier, as it no longer requires the user to
import an obscure type from the PAC.


hannobraun:
> This should also be changed in the usart example

Done. Had to rebase, as the example had changed in the meantime.

david-sawatzke:
LGTM!

PR #196, opened 2020-01-11
Add ctimer based pwm implementation

david-sawatzke:


david-sawatzke:
@hannobraun `RegProxy` really feels like it should have an unsafe `new` method, since I think you can otherwise create race conditions. Is there any reason it's not currently?

hannobraun:
Thank you, @david-sawatzke! I intend to take a look later today.



> `RegProxy` really feels like it should have an unsafe `new` method, since I think you can otherwise create race conditions. Is there any reason it's not currently?

I don't remember my reasoning back then in detail, but I think the key is this bit of documentation: https://github.com/lpc-rs/lpc8xx-hal/blob/ba40d07758b0504caf7844623c4aada47f5b8d1a/src/reg_proxy.rs#L48-L51

Basically, `RegProxy` doesn't enable anything you can't already do with raw registers.

The whole thing is a huge hack anyway. I still hope that https://github.com/rust-embedded/svd2rust/issues/213 will get implemented some day.

hannobraun:
Addendum to my comment earlier: I forgot that `RegProxy::new` doesn't have an argument. In its current form, it should definitely be `unsafe`. I think what I said earlier would make sense, if `RegProxy` took a reference to the register, just to prove the caller has access.

But yeah, all of it is a bit of a mess. I think it's fine, because it's just an internal helper, but I wouldn't mind making its constructor `unsafe`.

david-sawatzke:
> Suggestion: Rename UnconfiguredPwmPin to DetachedPwmPin and configure to attach. I believe that would be clearer

Thanks for the suggestion, it's more obvious what happens that way

src/ctimer.rs
```diff
@@ -0,0 +1,204 @@
...
+///
+/// Please refer to the [module documentation] for more information.
+///
+/// [`Peripherals`]: ../struct.Peripherals.html
+/// [module documentation]: index.html
+pub struct CTimer {
+    ct: CTIMER0,
+}
+
+/// An detached [`CTimerPwmPin`]
```
hannobraun:
```suggestion
/// A detached [`CTimerPwmPin`]
```

src/ctimer.rs
```diff
@@ -0,0 +1,204 @@
...
+///
+/// Use `attach` to assing an output to it
```
hannobraun:
```suggestion
/// Use `attach` to assign an output to it.
```

src/ctimer.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// prioritize it accordingly.
+    ///
+    /// [open an issue]: https://github.com/lpc-rs/lpc8xx-hal/issues
+    pub fn free(self) -> CTIMER0 {
+        self.ct
+    }
+}
+
+impl<CTOutput> DetachedPwmPin<CTOutput> {
+    /// Assings a pin to an `DetachedPwmPin`,
```
hannobraun:
```suggestion
    /// Assigns a pin to a `DetachedPwmPin`,
```

hannobraun:
Seems I don't have permissions to commit to your branch. You will have to apply the suggestions. Feel free to merge yourself, once you did that.

PR #201, opened 2020-01-14
Implement MRT abstraction

david-sawatzke:
TODO:
- Check what's up on the lpc82x

  The UM says it's a 31 bit timer, but the documentation for MODCFG.NOB suggests 24 bits
  
  @hannobraun Could you check? I unfortunately don't have an lpc82x device

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -121,6 +121,7 @@ pub mod dma;
 pub mod gpio;
 #[cfg(feature = "82x")]
 pub mod i2c;
+pub mod mrtimer;
 pub mod pmu;
 pub mod sleep;
 pub mod swm;
@@ -159,6 +160,7 @@ pub use self::dma::DMA;
 pub use self::gpio::GPIO;
 #[cfg(feature = "82x")]
 pub use self::i2c::I2C;
+pub use self::mrtimer::MRTimer;
 pub use self::pmu::PMU;
 pub use self::swm::SWM;
 pub use self::syscon::SYSCON;
@@ -220,6 +222,9 @@ pub struct Peripherals {
     /// I2C0-bus interface
     #[cfg(feature = "82x")]
     pub I2C0: I2C<init_state::Disabled>,
+
+    /// Multi-Rate Timer (MRT)
+    pub MRT0: MRTimer,
 
     /// Power Management Unit
     pub PMU: PMU,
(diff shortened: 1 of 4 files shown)


src/mrtimer.rs
```diff
@@ -0,0 +1,146 @@
...
+    reg_proxy::RegProxy,
+    syscon,
+};
+
+use embedded_hal::timer::{CountDown, Periodic};
+use nb::{Error, Result};
+use void::Void;
+
+/// Represent a MRT0 instance
+pub struct MRTimer {
```
hannobraun: I'd prefer the name `MRT` (and `mrt` for the module), as that makes is consistent with the user manual, making it clear what peripheral this API is for. It's also what we do for other peripherals.

src/mrtimer.rs
```diff
@@ -0,0 +1,146 @@
...
+    mrt: MRT0,
+}
+
+/// Represent a MRT0 channel
+pub struct MRTimerChannel {
```
hannobraun: I'd call this `MrtChannel`.

src/mrtimer.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// request already exists, comment on the existing issue, so we can
+    /// prioritize it accordingly.
+    ///
+    /// [open an issue]: https://github.com/lpc-rs/lpc8xx-hal/issues
+    pub fn free(self) -> MRT0 {
+        self.mrt
+    }
+}
+
+// TODO the lpc82x um isn't quite clear on how many bits the mrt has
```
hannobraun: The MRT has 31 bits. This has been wrong in the SVD, but has since been fixed (https://github.com/lpc-rs/lpc-pac/issues/18). The comment in the user manual is probably a leftover from whatever device they copy-pasted from.

src/mrtimer.rs
david-sawatzke: Sure, why not

src/lib.rs
```diff
@@ -160,7 +160,7 @@ pub use self::dma::DMA;
 pub use self::gpio::GPIO;
 #[cfg(feature = "82x")]
 pub use self::i2c::I2C;
-pub use self::mrtimer::MRTimer;
+pub use self::mrtimer::MRT;
```
hannobraun: Module is still called `mrtimer`, which is inconsistent with its contents. Please rename to `mrt`.

src/mrtimer.rs
david-sawatzke: Thanks for the confirmation

src/lib.rs
david-sawatzke: Should™ be fine now, after a careful check with `grep`

hannobraun:
And CI is green too. Thank you, @david-sawatzke!

PR #202, opened 2020-01-15
Add i2c to lpc845

david-sawatzke:
I2c port to lpc845

- This is completely untested
-  also didn't instantiate I2C{1-3} yet

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -118,18 +118,12 @@ where
         // default value (see user manual, sections 8.5.8 and 8.5.9).
 
         // Set I2C clock frequency
-        // Here's my thinking: The main clock runs at 12 Mhz by default. The
-        // minimum low and high times of SCL are set to 2 clock cyles each (see
-        // below), meaning a full SCL cycle should take 4 clock ticks. Therefore
-        // dividing the main clock by 8 (which is achieved by writing 7 below),
-        // should result in an I2C frequency near 400 kHz (375 kHz to be
-        // precise).
-        // None of that is correct, of course. When actually running, I'm
-        // measuring an SCL frequency of 79.6 kHz. I wish I knew why.
-        self.i2c.clkdiv.write(|w| unsafe { w.divval().bits(7) });
-
-        // SCL low and high times are left at their default values (two clock
-        // cycles each). See user manual, section 15.6.9.
+        // We just use the values for 400 kHz, from table 370 from the lpc845 um
+        self.i2c.clkdiv.write(|w| unsafe { w.divval().bits(5) });
+        self.i2c.msttime.write(|w| {
+            w.mstsclhigh().clocks_2();
+            w.mstscllow().clocks_3()
+        });
 
         // Enable master mode
         // Set all other configuration values to default.
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -97,7 +97,7 @@ where
     ///
     /// # Limitations
     ///
-    /// This method expects the I2C mode for PIO0_10 and PIO0_11 to be set to
+    /// This method expects the mode for SDA & SCL pins to be set to
     /// standard/fast mode. This is the default value.
     ///
     /// The I2C clock frequency is hardcoded to a specific value. For unknown
@@ -263,10 +263,10 @@ pub trait Instance:
     /// The interrupt that is triggered for this I2C peripheral
     const INTERRUPT: Interrupt;
 
-    /// The movable function that needs to be assigned to this I2C's Sda pin
+    /// The movable function that needs to be assigned to this I2C's SDA pin
     type Sda;
 
-    /// The movable function that needs to be assigned to this I2C's Scl pin
+    /// The movable function that needs to be assigned to this I2C's SCL pin
     type Scl;
 }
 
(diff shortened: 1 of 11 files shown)


david-sawatzke:
I also know why your timing is broken, the default for low & high cycles is 7, so 9 cycles

In the lpc845 um, there's also table 370 that recommends some values

hannobraun:
Sorry for the long delay. I'll take a look tomorrow.

src/i2c.rs
```diff
@@ -235,7 +249,54 @@ impl<State> I2C<State> {
...
+/// be considered breaking changes.
+pub trait Instance:
+    Deref<Target = pac::i2c0::RegisterBlock>
+    + syscon::ClockControl
+    + syscon::ResetControl
+{
+    /// The interrupt that is triggered for this I2C peripheral
+    const INTERRUPT: Interrupt;
+
+    /// The movable function that needs to be assigned to this I2C's Sda pin
```
hannobraun:
```suggestion
    /// The movable function that needs to be assigned to this I2C's SDA pin
```

src/i2c.rs
```diff
@@ -235,7 +249,54 @@ impl<State> I2C<State> {
...
+    type Sda;
+
+    /// The movable function that needs to be assigned to this I2C's Scl pin
```
hannobraun:
```suggestion
    /// The movable function that needs to be assigned to this I2C's SCL pin
```

src/i2c.rs
```diff
@@ -100,12 +105,12 @@ impl I2C<init_state::Disabled> {
...
     /// [`Enabled`]: ../init_state/struct.Enabled.html
-    pub fn enable(
+    pub fn enable<SdaPin, SclPin>(
         mut self,
         syscon: &mut syscon::Handle,
-        _: swm::Function<I2C0_SDA, swm::state::Assigned<PIO0_11>>,
-        _: swm::Function<I2C0_SCL, swm::state::Assigned<PIO0_10>>,
-    ) -> I2C<init_state::Enabled> {
+        _: swm::Function<I::Sda, swm::state::Assigned<SdaPin>>,
+        _: swm::Function<I::Scl, swm::state::Assigned<SclPin>>,
```
hannobraun: The documentation of this method still refers to PIO0_10 and PIO0_11.

src/i2c.rs
```diff
@@ -118,18 +118,12 @@ where
...
-        self.i2c.clkdiv.write(|w| unsafe { w.divval().bits(7) });
-
-        // SCL low and high times are left at their default values (two clock
-        // cycles each). See user manual, section 15.6.9.
+        // We just use the values for 400 kHz, from table 370 from the lpc845 um
+        self.i2c.clkdiv.write(|w| unsafe { w.divval().bits(5) });
+        self.i2c.msttime.write(|w| {
+            w.mstsclhigh().clocks_2();
+            w.mstscllow().clocks_3()
+        });
```
hannobraun: The method documentation needs to be updated. It should state that the frequency is hardcoded to 400 kHz, assuming the main clock is left at its default value.

david-sawatzke:
> It writes data just fine, but the clock is running at ~91 kHz.

@hannobraun Are you sure? Since the prescaler for the high/low times was changed from 7 to 2/3, this sounds wrong.

david-sawatzke:
@hannobraun I've added the port to the lpc845, but at the same time also needed to significantly revamp the clockconfig stuff, which im now fairly happy with (apart from possible improvements in the naming department).

I've tested usart, the old stuff & the new helper functions, but not i2c

hannobraun:
> Are you sure? Since the prescaler for the high/low times was changed from 7 to 2/3, this sounds very weird.

That's what the logic analyzer said :-)

It's weird, but I'm not surprised. It was weird before, and now we're using values from the LPC845 user manual and expect them to work on the LPC82x.

I'll take a look at the new commits and re-test I2C later today!

src/syscon/clocksource_845.rs
```diff
@@ -76,14 +56,108 @@ impl PeripheralClockSource for IOSC {
...
+            osrval,
+            _periphclock: PhantomData,
+        }
+    }
+}
+
+impl<PERIPH: crate::usart::Instance + PeripheralClockSelector>
+    UsartClock<(PERIPH, IOSC)>
+{
+    /// Create a new configuration with a specified baudrate
```
hannobraun: Should mention that this assumes a system clock of 12 MHz.

david-sawatzke:
@hannobraun The frequency fluctuates cause you're sampling too slow (remember the dots, luke)

I've goofed up the 400khz values, but it probably still should've worked. I've now tested some code  on the lpc845 without an attached i2c device with the latest commit and thus far it seems to work fine

hannobraun:
> The frequency fluctuates 'cause you're sampling too slow (remember the dots, luke)

Indeed I am. Looks like I haven't stopped being an idiot. Maybe I'll manage that next week :-)

> I've goofed up the 400khz values, but it probably should've worked. I've now tested some code on the lpc845 without an attached i2c device with the latest commit and thus far it seems to work fine

Nice! I'll re-test on LPC82x next week.

david-sawatzke:
I've changed some things to be more in line with the um, since it didn't quite work for me. I've tested it with the new example and it works great!

david-sawatzke:
Thanks for the review!

 Should be good to go now

PR #203, opened 2020-01-27
Adc impl

david-sawatzke:


examples/adc.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[entry]
+fn main() -> ! {
+    let p = Peripherals::take().unwrap();
+
+    let mut delay = Delay::new(p.SYST);
+    let swm = p.SWM.split();
+    let mut syscon = p.SYSCON.split();
+
+    #[cfg(feature = "82x")]
```
hannobraun: I think we don't need LPC82x support for this. Regressions would be bad, of course, but we don't need feature parity on new features. I don't have time to really get into it right now, and I think it's perfectly fine to leave this until someone (maybe me, later) actually needs it, and invests the time to get it up to par.

(By the way, this won't work anyway, as the example requires `845` in Cargo.toml.)

src/adc.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+impl<PIN> OneShot<ADC, u16, PIN> for ADC
+where
+    PIN: Channel<ADC, ID = u8>,
+{
+    type Error = ();
+
+    // TODO Nonblocking
```
hannobraun: Suggestion: Remove this TODO, get this merged, leave further work to follow-up PRs.

src/syscon.rs
```diff
@@ -15,16 +15,7 @@ pub mod frg;
...
-/// Clock configuration for peripherals
-pub mod clocksource_82x;
-#[cfg(feature = "82x")]
-pub use clocksource_82x as clocksource;
-#[cfg(feature = "845")]
-/// Clock configuration for peripherals
-pub mod clocksource_845;
-#[cfg(feature = "845")]
-pub use clocksource_845 as clocksource;
+pub mod clocksource;
```
hannobraun: Nice!

Though I'm slightly confused that it works. I don't see any updates to the examples (and I assume this should break some of them). But CI is still green?

src/adc.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// This method is only available, if `ADC` is in the [`Disabled`] state.
+    /// Code that attempts to call this method when the peripheral is already
+    /// enabled will not compile.
+    ///
+    /// Consumes this instance of `ADC` and returns another instance that has
+    /// its `State` type parameter set to [`Enabled`].
+    ///
+    /// # Limitations
+    ///
+    /// This assumes that the clock is running at 500 kHz for the calibration
```
hannobraun: Might be worth specifying which clock that is. At least it's not obvious to me, not having touched an LPC800 ADC in a while.

src/adc.rs
```diff
@@ -1,4 +1,34 @@
-//! TODO
+//! API for ADC
+//!
+//! # Examples
+//!
+//! Read a single value
```
hannobraun:
```suggestion
//! Read a single value:
```

src/adc.rs
david-sawatzke: I think it's fine to remove this, since it's already documented in clocksource.rs, which is more clear

examples/adc.rs
david-sawatzke: Yeah, wanted ci to pass. But it's obvious enough how to adapt it to lpc82x

src/syscon.rs
david-sawatzke: In clocksource.rs I've added `pub use clocksource_845::*;`, which makes it mostly non-breaking

hannobraun: I saw that, but I somehow had missed that these were already used `as clocksource` in `syscon` (yes, I see now that it's right here in the diff). Anyway, all makes sense now.

src/adc.rs
```diff
@@ -152,9 +148,8 @@ where
 {
     type Error = ();
 
-    // TODO Nonblocking
     fn read(&mut self, _: &mut PIN) -> nb::Result<u16, Self::Error> {
-        // Start reading ofthe channel
+        // Start the measurment of the given channel
```
hannobraun:
```suggestion
        // Start the measurement of the given channel
```

src/adc.rs
```diff
@@ -22,7 +22,7 @@
 //!     .assign(swm.pins.pio0_7.into_swm_pin(), &mut swm.handle);
 //!
 //! // Read a single value
-//! let adc_value = block! {adc.read(&mut adc_pin)}
+//! let adc_value = block!{adc.read(&mut adc_pin)}
```
hannobraun:
```suggestion
//! let adc_value = block!(adc.read(&mut adc_pin))
```

Never saw the `block!` macro being used with curly braces. Shouldn't make a difference, but this should reduce the chance of confusion for anyone new to Rust.

PR #204, opened 2020-01-28
Spi impl

david-sawatzke:


examples/spi_apa102.rs
```diff
@@ -0,0 +1,85 @@
...
+    let mut spi = p.SPI0.enable(
+        &spi_clock,
+        &mut syscon.handle,
+        MODE,
+        spi0_sck,
+        spi0_mosi,
+        spi0_miso,
+    );
+
+    // We're done. Let's do nothing until someone resets the microcontroller.
```
hannobraun: Looks like a copy-pasted remainder that doesn't fit here.

src/spi.rs
```diff
@@ -0,0 +1,249 @@
...
+
+use crate::{
+    init_state, pac,
+    swm::{self, FunctionTrait, PinTrait},
+    syscon::{self, clocksource::SpiClock, PeripheralClock},
+};
+
+/// Interface to a SPI peripheral
+///
+/// Controls the SPI.  Use [`Peripherals`] to gain access to an instance of
```
hannobraun:
```suggestion
/// Controls the SPI. Use [`Peripherals`] to gain access to an instance of
```

david-sawatzke:
Ok, fixed the review issues and some other things (pins overlap with i2c, bitlength was at 1) as well as testing it, ready to go now

PR #208, opened 2020-02-06
Add method to USART that clears NVIC pending flag

hannobraun:


david-sawatzke:
The description seems a bit confusing to me. At first I understood it to mean that you have to call it at the end of every interrupt, which doesn't seem to be the case with the c example code & drivers.

My understanding is (and feel free to correct me if I'm completely wrong) that if a flag is set in stat & enabled in intenset, it asserts the interrupt line and deasserts it if the flag is cleared. The only scenario where it makes sense to clear the pending flag is if the usart interrupt is blocked and the flag is then cleared outside the usart interrupt. 

Maybe something like this:
```
This needs to be called after an usart interrupt flag 
has been handled outside the usart interrupt. 

Otherwise, the usart interrupt will fire again immediately.
```

LGTM otherwise

hannobraun:
Okay, this is interesting. I clearly remember cases where interrupt handlers never stopped re-running until I cleared the pending flag. And the case that prompted this pull request was a main loop that turned into a busy loop after the first interrupt (`wfi` never blocked again). I thought that clearing NVIC was expected and normal.

But some research shows that indeed you are right, and that I somehow managed to hold on to this years-old misconception until now :-)

I wonder what's happening in my current case. I only have the RXRDY interrupt enabled, and according to the documentation, the flag should be cleared when reading RXDAT or RXDATSTAT. I'll have to take a closer look.

In any case, I'll update the documentation later.

hannobraun:
Updated the documentation. I opted for something that explains what the method does, instead of trying to describe when it might be good to use it.

PR #210, opened 2020-02-14
Improve split between USART's receiver and transmitter

hannobraun:


hannobraun:
Please don't merge. I'm having problems downstream, possibly related to receiving. I'll investigate.

hannobraun:
Okay, whatever my problem is, I don't think it's related to this pull request. Now everything works again. Since the whole thing happened at the same time GDB/OpenOCD were acting up, I'm inclined to blame those two.

I'll keep an eye out, but for now, I don't think there's anything wrong with this pull request. Feel free to review/merge.

david-sawatzke:
I *really* dislike the magic numbers for the TXDAT address. 
What do you think about using the REGISTER_BLOCK in the `Instance`:
```
const REGISTER_BLOCK: *const pac::usart0::RegisterBlock = <pac::$instance>::ptr();
```
and getting the address itself inside `end_addr`:
```
unsafe { &(*I::REGISTER_BLOCK).txdat as *const _ as *mut u8 }
```
?

With that, you could remove all other additions to the trait, which would simplify the code quite a bit and be worth it. (In my opinion, certainly not a blocker)

LGTM otherwise

timokroeger:
I also use the method described by david in the EFM32PG12-HAL.
Actually all that is required is this [trait with ptr() method](https://github.com/timokroeger/efm32pg12-hal/blob/4ef520c84f0b42998a399c7a8900a5ec1c9f9632/src/util.rs#L6) which acts as super trait for the peripheral instance.

hannobraun:
Thank you for the suggestions, both of you. I've pushed a new commit that gets rid of the magic numbers.

@david-sawatzke 
> With that, you could remove all other additions to the trait, which would simplify the code quite a bit and be worth it. (In my opinion, certainly not a blocker)

I might be misunderstanding what you're saying, but I think the other additions to the trait are required because of `RegProxy`. At least I can't think of a way to get rid of them, without getting rid of `RegProxy`.

We could just get rid of `RegProxy`, of course, but then we'd need an `unsafe` block everywhere we're using a register. As far as I can tell, this is what all other HALs are doing. Maybe we should just do that, as the safety of `RegProxy` is dubious anyway (the constructor should be `unsafe`).

`RegProxy` is useful to "move" registers somewhere else, but in cases where multiple APIs use the same registers (like here), it probably doesn't provide and advantage. In fact, it's probably more error-prone than `unsafe` everywhere, as it hides what's going on, and what's required to keep everything sound.

But in any case, I'd prefer to merge this pull request as-is and continue this discussion elsewhere (I've opened https://github.com/lpc-rs/lpc8xx-hal/issues/211). Should we decide to get rid of `RegProxy`, I'm happy to do the leg-work of removing it and simplifying the code currently using it.

PR #213, opened 2020-02-25
Make dependency on Termion optional

### LINKED ISSUE #212: Build broken on windows because of `termion` crate
ref: https://gitlab.redox-os.org/redox-os/termion/issues/167
Issue can be fixed by reverting 6638553eb03d42883a95ba56426aff7271267503

hannobraun:
It doesn't build on Windows, which I think is enough reason to make it
optional and unselected by default.

I don't think having it as an optional dependency is particularly
useful, but I didn't want to remove it outright. I'm hoping that
Termion's problems on Windows are temporary, and once it is fixed,
restoring the previous functionality will be easy.

Close #212

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,8 +46,11 @@ panic-halt    = "0.2.0"
 cortex-m-rtfm = "0.5.1"
 
 
-[build-dependencies]
-termion = "1.5.3"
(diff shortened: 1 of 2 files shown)


hannobraun:
@timokroeger, it would be great if you could verify that this fixes your problem.

PR #215, opened 2020-02-25
Rename `Peripherals` to `Device`

hannobraun:
Based on #214. I could have added this as part of #214, but I figured it might be a more controversial change, and thus worth submitting separately.

### DIFF
--- a/examples/adc.rs
+++ b/examples/adc.rs
@@ -10,40 +10,41 @@ use nb::block;
 
 use lpc8xx_hal::{
     cortex_m_rt::entry, delay::Delay, prelude::*,
-    syscon::clock_source::AdcClock, usart, CorePeripherals, Peripherals,
+    syscon::clock_source::AdcClock, usart, CorePeripherals, Device,
 };
 
 #[entry]
 fn main() -> ! {
     let cp = CorePeripherals::take().unwrap();
-    let p = Peripherals::take().unwrap();
+    let device = Device::take().unwrap();
 
     let mut delay = Delay::new(cp.SYST);
-    let swm = p.SWM.split();
-    let mut syscon = p.SYSCON.split();
+    let swm = device.SWM.split();
+    let mut syscon = device.SYSCON.split();
 
     let mut handle = swm.handle.enable(&mut syscon.handle); // SWM isn't enabled by default on LPC845.
 
     // Set baud rate to 115200 baud
     let clock_config = usart::Clock::new_with_baudrate(115200);
 
-    let tx_pin = p.pins.pio0_25.into_swm_pin();
-    let rx_pin = p.pins.pio0_24.into_swm_pin();
+    let tx_pin = device.pins.pio0_25.into_swm_pin();
+    let rx_pin = device.pins.pio0_24.into_swm_pin();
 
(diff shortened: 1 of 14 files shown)


hannobraun:
Rebased and ready for review/merge.

david-sawatzke:
Sorry for letting this sit for a bit, but I wanted to have some time to look at this a bit more in-depth before replying.

I agree that `Peripherals` may not be an ideal name after moving pins, but I think (currently) it's the best choice:
- I disagree with the reasoning that it improves consistency. It's a wrapper around the `svd2rust` `Peripherals` API, with the same API. The `device` field in rtfm also contains the `Peripherals` type. If (almost?) every other hal uses the svd2rust generated `Peripherals` directly without wrapping it, I think it's more consistent to stick with the name. This (for me) is the major reason why I think keeping the name is the best choice.

- Whilst with `Peripherals` one can argue over how accurate the name is, `Device` also implies that it contains everything, while it's missing the core peripherals (to be a smart-ass & cover all my bases).

If you disagree, I'd be alright with this being merged, since there doesn't seem to be a clear-cut "best" option.

hannobraun:
> Sorry for letting this sit for a bit, but I wanted to have some time to look at this a bit more in-depth before replying.

No worries. Always feel free to take as much time as you need!


> I disagree with the reasoning that it improves consistency. [...]

I was thinking about RTFM and not the other HALs, because RTFM might be used together with this HAL, in the same project. But you're right, the consistency argument is a weak one.


> Whilst with `Peripherals` one can argue over how accurate the name is, `Device` also implies that it contains everything, while it's missing the core peripherals (to be a smart-ass & cover all my bases).

Hmm, I'd actually say your argument favors `Device`. `Peripherals` also doesn't contain the core peripherals, and given its name, that's an even worse inaccuracy than in the case of `Device`. :-)


> If you disagree, I'd be alright with this being merged, since there doesn't seem to be a clear-cut "best" option.

I agree that there's no clear-cut best option here. I still favor `Device`, as it's shorter. Case in point, I think 99% of the HAL examples I've seen use `p` or `dp` (for device peripherals, I presume) as the variable name for `Peripherals`.

I think our examples look better with `core` and `device`, as those names are short enough to not be a burden, but long and different enough to be clearly distinguished (as opposed to `cp`/`p` or `cp`/`dp`).

---

I'm not sure how to proceed. I'm clearly in favor of `Device`, but don't feel so strongly about the matter that I just want to overrule you. You say you're alright with merging this, but I'd feel bad about winning just for being more stubborn :-)

If you happen to feel stubborn yourself while reading this, feel free to just close. If not, maybe let's leave it open for a while to reflect further. It's an easy change, so merge conflicts won't be much of an issue.

hannobraun:
- Rebased
- Fixed typo in commit message

hannobraun:
Rebased.

hannobraun:
Having thought this over for a while, I still prefer `Device`, for the reasons mentioned. I still don't feel that strongly about this (especially given that there's no clear-cut "best" solution), and don't think it's worth having this pull request open.

Closing for now. Maybe I can sneak it by you at a later date :-)

PR #216, opened 2020-02-26
Clean up `pins` module

hannobraun:
I'm still working on pin/GPIO improvements. This seemed like a good self-contained PR to make, before I get into the more heavy-handed changes.


hannobraun:
Pushed two new commits that fix the documentation problems you mentioned, as well as some others. For what it's worth, I plan to do a full review of the documentation within the next few months.

PR #217, opened 2020-02-27
Clean up and simplify GPIO API

hannobraun:
This is another milestone in my ongoing effort to update the pins/GPIO API. It's based on #216, and I suggest reviewing/merging that one first.

PR #218, opened 2020-02-27
Simplify `GPIO`/`SWM` constructors

hannobraun:
Still not what I set out to do with the pin/GPIO API, but I keep finding things that can be improved on the way.

This PR is based on #217. I recommend reviewing/merging that one first.


hannobraun:
Rebased, ready for review.

PR #219, opened 2020-03-02
Don't require `GpioPin` to borrow GPIO peripheral

hannobraun:


david-sawatzke:
The need to add tokens for each pin seems really involved to me, for negligible gain.

So with this change, as soon as one pin is used as an output, you can't change the gpio state (since I can't see a way to get the token back). The only reason one would want to disable the gpio peripheral without using it once is on the lpc82x, where it's on by default.

I can think of two ways to do this that seem better to me:
- Add some functionality to "freeze" the gpio peripheral
- Make disabling the peripheral unsafe
  - Only in a small number of applications is the gpio peripheral never used
  - It also makes it possible to temporarily disable the gpio in some sleep state

hannobraun:
Well, I totally disagree :-)

> The need to add tokens for each pin seems really involved to me, for negligible gain.

It is more involved than what we had before, but what we had before was shit. This is not shit, in my opinion, which makes the gain quite significant.

The previous solution, having each GPIO pin borrow the GPIO peripheral, only worked well in the most basic cases. As soon as you were moving a pin somewhere else, you needed explicit lifetimes. This can already be tedious by itself, but in the context of embedded development, this means that the lifetime usually needs to be `'static`, which means `GPIO` needs to life in a `static`.

Yes, the tokens add complexity on the implementation side, and they make the general case slightly more verbose. I strongly believe that this is worth it, compared to either having to employ the `static` workaround (which is way more involved than having to provide a token), or not having any protection at all, against accidental disabling.

> So with this change, as soon as one pin is used as an output, you can't change the gpio state (since I can't see a way to get the token back). The only reason one would want to disable the gpio peripheral without using it once is on the lpc82x, where it's on by default.

This is false. You need a token to put the pin into the GPIO state (`Pin::into_input_pin`/`Pin::into_input_pin`). You _do not_ need a token to change between input and output after that (`GpioPin::into_input`/`GpioPin::into_output`).

What is true, is that there's no way to get the token back from the `GpioPin`, so it's not possible to disable the GPIO peripheral after having used it once. This is trivial to fix. I didn't add it, because I currently have no need for it, and because I was under the (mistaken) impression that this already wasn't possible with the old API. I'm happy to add this method, if you prefer to have it right now.

> I can think of two ways to do this that seem better to me:
> 
> * Add some functionality to "freeze" the gpio peripheral

This is effectively what this PR does. While freezing the GPIO peripheral could probably be done more simply, this solution is just one trivial step away from supporting disabling the GPIO peripheral, once all pins are no longer in use. I can't think of a simpler design that would still be able to do that, while providing the same correctness guarantees.

> * Make disabling the peripheral unsafe
>       
>   * Only in a small number of applications is the gpio peripheral never used
>   * It also makes it possible to temporarily disable the gpio in some sleep state

My ambition for this HAL has always been to provide an API that guarantees correctness, while not limiting the user:
- Correctness is important because, in my experience, every mistake that can be made will be made, sooner or later.
- Flexibility is important, because every time you limit what the user can do with the HAL, compared to what they could do without it, you force some section of your user base to use the PAC. And the users who need the flexibility likely have non-trivial applications and need correctness guarantees the most.

I think it's totally fine to write APIs that don't fully guarantee correctness and/or limit the user,  if a better solution is not practical or available. In this case, a correct and flexible solution is practical (as demonstrated by this PR), and already mostly implemented (minus the one method that transforms `GpioPin` into `Pin` and gives the token back).

---

There's a trade-off to be made between simplicity, flexibility, and correctness. If we pick flexibility and correctness, we can't always have simplicity. I think it would be totally worthwhile to have an API that is simple and correct (which necessarily means being opinionated and limited). But such an API can be built on top of the flexible/correct one, while the other way around is not possible.

david-sawatzke:
Sorry, seems what I previously wrote didn't really make it clear what I was referring to & suffered from overly lossy compression of thoughts.

> It is more involved than what we had before, but what we had before was shit. This is not shit, in my opinion, which makes the gain quite significant.

This is way better than the previous solution and I'd rather have this than the previous solution. I was mostly thinking to this vs. the alternate solutions. 

> This is false. You need a token to put the pin into the GPIO state (Pin::into_input_pin/Pin::into_input_pin). You do not need a token to change between input and output after that (GpioPin::into_input/GpioPin::into_output).

I was referring to the state of the gpio peripheral, but your next paragraph already covers that.

---- 
Regarding the trade-off in general:
- Of course, correctness in safe code should be *paramount*. 
- Providing full flexibility often isn't possible in safe abstractions on MCUs, see (for example) dmas with non-static buffers
  (which, now that I think about it, is the major benefit of safe abstractions)
- When more flexibility in safe code comes at a significant cost, it's sometimes better to make (client & lib) code simpler by allowing a bit of `unsafe`ty when it's only for a few corner cases 
  - The `unsafe` calls should also be pretty obvious in what they do and what the effects are (a huge issue with PACs)

My belief is that disabling the gpio peripheral  is a *very* rare occurrence and I can only think of three scenarios where it may be done:
- Disabling the gpio peripheral right away (only on lpc82x)
- Doing some small init & then disabling the gpio peripheral (for example, briefly resetting a bus on startup)
- Having a deep sleep that's controlled with a timer & disable the gpio while sleeping
  - It seems that most people wouldn't go the extra mile to gather all the pins & enable them again afterwards. If I were in this situation probably just disable the gpio peripheral directly (via pac).

In this specific case, the flexibility is so rarely needed that I don't think the addition of proxy objects and the need to carry them around for every gpio pin is worth it. Furthermore, in all non-trivial cases (mostly the third option), it seems significantly harder use this api, since you need to re-gather all pins (if not nigh impossible due to dependencies), especially without overhead (since every gpio pin needs to be configured again afterwards)

Thus, in this specific case, I believe that making `disable` unsafe and otherwise mostly just removing the `'gpio` lifetime references, is the best choice. It makes the api more flexible and simpler, at no cost to correctness, while still allowing one to make a layer that makes disabling the gpio safe. It comes at the cost of something else, the creation of more `unsafety`ness (very little though). 

---------
Since this is admittedly a significant improvement over the status quo, I'd merge this PR now if you're not totally convinced by this (which I doubt ;-) ) and if this irks me enough in the next few months I'll have a go at it.

hannobraun:
Thank you for clarifying your position, David. I think we agree on the facts now, and just disagree on how to weigh their relative importance (which is always a great point to reach in a discussion).

I have to say, you're making good points, and I'm starting to come around to your position. My desire to provide a 100% correct API might just be irrational, in this case.

That said, I'd still like to take you up on your offer to merge this now and revisit later, for the following reasons:
- We agree that this is an improvement over what we already have.
- I have in-progress code (in [another project](https://github.com/braun-embedded/lpc845-test-stand)) that will benefit from this being merged.
- I have more urgent work, so having to revisit this code right now wouldn't be my most favorite thing.

I've made a note to revisit this design at a later date. Quite possible that I'll be fully convinced at that point. In the meantime, I'd be happy to see pull requests with the aim at improving this code.

PR #223, opened 2020-03-16
Don't tie USART clock to specific USART instance; cleanups

hannobraun:
Based on #222. I recommend reviewing/merging that one first.

Originally I just meant to remove the connection between a USART clock a specific USART instance, as that's not necessary, and prevents re-use of configuration between multiple USART instances (which is a bit annoying). It turned into a meandering cleanup operation of USART clock code.


hannobraun:
Rebased, ready for review.

PR #224, opened 2020-03-19
Clean up `pins` module

hannobraun:


src/pins/state.rs
```diff
@@ -0,0 +1,67 @@
+//! Contains types that indicate pin states
+//!
+//! Please refer to [`Pin`] for documentation about how these types are used.
```
david-sawatzke: `Pin` isn't linked (compared to all the other links to `Pin` in this file)

src/pins/state.rs
```diff
@@ -0,0 +1,67 @@
...
+use core::marker::PhantomData;
+
+/// Implemented by types that indicate pin state
+///
+/// [`Pin`] uses this type as a trait bound for the type parameter that
+/// indicates the pin's state. This is done for the purpose of
+/// documentation, to show which states a pin can be in. Other than that,
+/// this trait should not be relevant to users of this crate.
+///
+/// [`Pin`]: ../struct.Pin.html
```
david-sawatzke: I think this link isn't valid anymore

david-sawatzke: (also applies to all the other links to `Pin`)

hannobraun: It is actually, due to the public re-export in `pins`. That's what I think anyway, but the link does work in the generated output.

hannobraun:
I fixed all broken links in the `pins` documentation and updated some of it. Merging, since the previous commits were already approved, and the new commits only change documentation, which is largely outdated and in need of an update anyways.

PR #226, opened 2020-03-20
Make `GpioPin` `Send`

hannobraun:
`GpioPin` wasn't `Send` before, since `Registers` contains
`UnsafeCell`s, which aren't `Send`. This prevented moving `GpioPin`
instances into an interrupt handler using RTFM.

I think we could just implement `Send` for `GpioPin` instead, and even
though I think that should be fine, I don't fully understand the
implications. I opted to just remove `Registers` as a field and
construct it on demand, which is similar to what we do for other
peripherals.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -171,7 +171,6 @@ impl GPIO<init_state::Enabled> {
 /// A pin used for general purpose I/O (GPIO)
 pub struct GpioPin<T, D> {
     token: pins::Token<T, init_state::Enabled>,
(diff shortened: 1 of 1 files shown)


hannobraun:
Rebased. Not sure why GitHub was complaining. There weren't any conflicts to resolve.

Merging, since the change were already approved. Travis fails on the nightly channel, but the others pass, and the nightly failure is unrelated to this PR (rustfmt can not be installed).

PR #227, opened 2020-03-20
Add PININT API

hannobraun:
This pull request is based on #226. I recommend reviewing/merging that one first.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -118,6 +118,8 @@ pub mod dma;
 pub mod gpio;
 pub mod i2c;
 pub mod mrt;
+#[cfg(feature = "845")]
+pub mod pinint;
 pub mod pins;
 pub mod pmu;
 pub mod sleep;
@@ -159,6 +161,8 @@ pub use self::dma::DMA;
 pub use self::gpio::GPIO;
 pub use self::i2c::I2C;
 pub use self::mrt::MRT;
+#[cfg(feature = "845")]
+pub use self::pinint::PININT;
 pub use self::pmu::PMU;
 pub use self::spi::SPI;
 pub use self::swm::SWM;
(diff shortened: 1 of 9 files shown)


hannobraun:
Rebased, ready for review.

src/pinint/traits.rs
```diff
@@ -0,0 +1,16 @@
+/// Implemented by types that identify pin interrupts
+///
+/// This trait is an internal implementation detail and should neither be
+/// implemented nor used outside of LPC8xx HAL. Any changes to this trait won't
+/// be considered breaking changes.
+pub trait Trait {
```
david-sawatzke: We should probably look into sealed traits for the future (Should probably not be fixed in this PR, to have a consistent style)

src/lib.rs
```diff
@@ -375,10 +383,17 @@ pub struct Peripherals {
 
     /// Pin interrupt and pattern match engine
     ///
-    /// A HAL API for this peripheral has not been implemented yet. In the
-    /// meantime, this field provides you with the raw register mappings, which
-    /// allow you full, unprotected access to the peripheral.
-    pub PINT: pac::PINT,
+    /// A HAL API for this peripheral has not been implemented yet for LPC8x. In
```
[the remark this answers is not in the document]
hannobraun: Yes, thanks.

Let me try something:
```suggestion
    /// A HAL API for this peripheral has not been implemented yet for LPC82x. In
```

src/pinint/traits.rs
hannobraun: Yes, definitely.

I have been aware of that technique for a while, and the reason I haven't done it for the newer traits is just pure laziness (which isn't a valid reason for anything). I'll keep it in mind for future traits, and if I happen to do more cleanup operations, I'll fix older traits to, as I go.

hannobraun:
Review comment applied. Merging.

PR #230, opened 2020-03-30
Improve MRT API

hannobraun:


hannobraun:
The Travis failure is only on nightly and seems unrelated to this PR. Stable and beta run fine.

hannobraun:
> I don't see a reason to duplicate the trait methods here. There doesn't seem to be anything that makes this case different from other rust traits

I think the `embedded-hal` traits are different from many other traits, in that they define the core API of the type that implements them. You are just as likely (or more) to use the trait methods with the concrete type, as you are to use them through the trait only.

This is different from many other traits, which often provide some extra thing that makes use of the type more flexible in generic situations.

In addition, since this is the core API of our type, we might decide to deviate from what the trait defines. As an example, I've already been in situations where I thought, "wouldn't it be nice to know that this timer is ready to be configured, and not about to fire an interrupt". If we added type state to the timer API for that reason, we'd have to implement the `embedded-hal` trait through some extra type (`timer.into_flexible_timer`, or something like that).

This last point isn't relevant to this PR, but it provides some background on why I like the idea of providing a full core API as inherent methods and implementing the `embedded-hal` traits as an extra thing. It gives us flexibility in implementing the API that makes the most sense for us.

> it might lead to confusion, since two functions with the same name exist.

I think the potential for confusion is minimal, and any confusion that might exist can be alleviated by documentation. Confusion would be even further reduced, if this became "the way" we implemented `embedded-hal` traits in this crate, and maybe beyond.

In any case, I think the potential for confusion is outweighed by the very real confusion of "how do I use this?", because the important methods are hidden further down in the documentation.

> (I'm also wondering why the rust compiler doesn't warn you when the trait is in scope & the method is used, but that's another issue)

I'm no expert on language semantics, but I would assume that inherent methods always have priority. This makes sense to me, because you might want to implement a trait for a type you don't control (so you might not be able to avoid naming collisions). And if both are in scope, the trait method can always be called using the explicit `<MyStruct as Trait>::method(my_value)` syntax.

---

I think I'm right on this, but I don't feel that strongly about this particular case. If you're not convinced by my arguments, let me know, I'll just drop the commit.

david-sawatzke:
> This is different from many other traits, which often provide some extra thing that makes use of the type more flexible in generic situations.

You're right that many traits handle extensions, but in quite a few cases the extensions is the core api (see for example iterators).

> In addition, since this is the core API of our type, we might decide to deviate from what the trait defines.

In that case it'd make sense, but the trait methods would (probably) differ a bit from the inherent methods.

> I would assume that inherent methods always have priority.

Yeah, tested it and the inherent one takes priority.

-----

I'd prefer to drop this, duplicate methods just feel a bit hacky. The rust docs just link to the "main" trait in the description in similiar situations, see [std::iter::Chain](https://doc.rust-lang.org/std/iter/struct.Chain.html), that could at least alleviate the concerns in the documentation department.

hannobraun:
Understood. I'll update the pull request later.

One more reply though:
> You're right that many traits handle extensions, but in quite a few cases the extensions is the core api (see for example iterators).

I think iterators are yet another case, where all you care about is the trait, and the concrete type is just an implementation detail. If the current feature set of Rust weren't limited in that regard, you could just as well type `impl Iterator` everywhere you use an iterator type.

hannobraun:
Dropped the commit introducing the inherent methods and fixed the resulting conflicts. Should be good to go now.

PR #231, opened 2020-04-03
Make doc tests compile

hannobraun:


hannobraun:
CI for nightly channel fails because `rustfmt` is not available. Seems irrelevant. The other channel builds succeed.

PR #235, opened 2020-05-06
Improve GDB/OpenOCD configuration

hannobraun:


openocd_84x.cfg
```diff
@@ -1,5 +1,3 @@
-# You might need to reset the chip after loading a new program
```
david-sawatzke, line 1: I'd like to keep this comment, since some people (me) keep gdb open and simply load the program again after recompiling, and this issue has bitten me more than once (after forgetting that it exists, of course).

hannobraun: Sounds reasonable. I just re-run `cargo run` every time and hadn't considered your use case. I'll drop the commit later.

hannobraun:
Dropped the undesired commit. Merging, since the rest was approved.

PR #239, opened 2020-05-25
Finish my current round of clock source clean-up

hannobraun:
The clock source code is not perfect by any means, but I think this PR leaves it in a good enough state to focus on other stuff for a while.


hannobraun:
The Travis build fails on nightly only, because `rustfmt` isn't available. Stable and beta build fine.

PR #240, opened 2020-05-26
Add support for I2C errors and interrupts

hannobraun:


src/i2c/peripheral.rs
```diff
@@ -209,3 +224,62 @@ where
...
+        // Check for error flags. If one is set, clear it and return the error.
+        if stat.mstarbloss().bit_is_set() {
+            i2c.stat.write(|w| w.mstarbloss().set_bit());
+            return Some(Self::MasterArbitrationLoss);
+        }
+        if stat.mstststperr().bit_is_set() {
+            i2c.stat.write(|w| w.mstststperr().set_bit());
+            return Some(Self::MasterStartStopError);
+        }
+        if stat.monov().bit_is_set() {
```
david-sawatzke, line 303: Not sure if it makes sense to include this, since it's only relevant in monitor mode and we don't check the slave flags as well. (not particularly important)

src/i2c/peripheral.rs
```diff
@@ -99,6 +99,23 @@ where
...
 
+impl<I> I2C<I, init_state::Enabled>
+where
+    I: Instance,
+{
+    /// Read and clear a detected error
+    ///
+    /// The read and write method will return an error, if one was detected.
+    /// However, only one error will be returned, even though, theoretically,
+    /// multiple could have been detected.
```
david-sawatzke, line 126:
```suggestion
    /// The `read` and `write` methods will return an error and clear it, if one was detected.
    /// However, if multiple errors occur, only one error will be returned and cleared.
```

src/i2c/peripheral.rs
hannobraun: I figured it's better to check a useless flag now, then to forget to add the check later. I thought I had added checks for all the slave error conditions as well, but I'll double-check later.

I don't feel strongly about this, but since you don't seem to feel strongly either, I'm inclined to leave the checks in.

david-sawatzke: Just checked, all slave error conditions are handled. Probably misread some of the slave non-error conditions.

PR #243, opened 2020-06-17
Migrate from RTFM to RTIC

hannobraun:
The RTFM framework has been renamed to RTIC. Everything else should be
the same, so this change should basically be a no-op.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ version      = "0.3.0"
 
 [dev-dependencies]
 panic-halt    = "0.2.0"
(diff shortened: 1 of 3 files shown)


hannobraun:
I don't know why there are two CI builds (at least two are displayed for me here). The first shows as failing. This is only the case on nightly, for unrelated reasons. Stable and beta are fine. The second currently display as being in progress, but actually has the same result as the first one.

So yeah, everything's fine. As far as CI is concerned, this is ready to be merged.

david-sawatzke:
I suppose travis is just being weird. The pr looks good

PR #244, opened 2020-06-18
Release version 0.7.0

hannobraun:
It's been a while since we published the last release, and I think the next one is overdue. As indicated by the changelog, it's easily going to be our biggest release yet. The reason I want to get it out right now, is because of the aforementioned article I'm writing. I don't want to depend on an outdated version, and I also don't want to use a GitHub dependency, if I can avoid it.

@david-sawatzke: Please let me know what you think. Don't merge this yet, as I still have to do a bunch of steps once you've approved this:

- [x] Update version in README
- [x] Update the date in the changelog
- [x] Publish on crates.io
- [ ] Merge this pull request
- [ ] Create and push the release tag

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,88 @@
 
+<a name="v0.7.0"></a>
+## v0.7.0 (2020-06-22)
+
+- Improve documentation ([#180], [#200], [#231], [#233], [#234])
+- Improve `cargo run` configuration ([#181], [#235])
+- Enforce consistent formatting with rustfmt ([#184], [#186])
+- Fix and improve build script ([#185], [#187], [#193], [#194], [#197], [#213])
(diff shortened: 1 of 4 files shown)


hannobraun:
Travis fails on nightly, for the usual reasons. Stable and beta are green.

hannobraun:
Thanks for having a look! I'll do the release soon-ish, likely Monday at the latest.

> Seems good, this is already a bit much for a single release. Great work on the changelog, might be a wise idea to maintain an "unreleased" changelog in the future.

Thank you.

I'd guess maintaining an ongoing changelog is going to be a hassle to stay on top of, and also more error-prone. I'd be willing to try it, if you want to, but I think the current approach is fine.

hannobraun:
All done!

PR #245, opened 2020-06-19
Avoid spurious CI failures by only installing rustfmt when needed

david-sawatzke:
This is pretty annoying, see https://github.com/lpc-rs/lpc8xx-hal/pull/244#issuecomment-646547406


hannobraun:
This is great, thank you!

PR #253, opened 2020-07-10
Add support for I2C slave mode

hannobraun:
This pull request is based on #252. I recommend reviewing/merging that one first.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -130,6 +130,10 @@ name              = "gpio_timer"
 required-features = ["rt-selected"]
 
 [[example]]
+name              = "i2c_master_slave"
+required-features = ["rt-selected", "845"]
+
+[[example]]
 name              = "i2c_vl53l0x"
 required-features = ["rt-selected", "82x"]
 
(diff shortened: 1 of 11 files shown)


david-sawatzke:
Haven't looked at this in-depth, but looks good to me! -> approved

hannobraun:
Okay, so here's the thing: When I submitted this, it worked. I verified this with the provided example, using RTT and a logic analyzer. It no longer works for me now, on none of my boards.

Some context: I've been having trouble getting slave mode to work for a while. After much investigation, I concluded that one of the boards I was using was likely damaged. I was using [a test setup](https://github.com/braun-embedded/lpc845-test-stand) with two boards, one acting as master, one as slave. Since I didn't have a replacement at hand, I implemented this new design and tested successfully on the other one.

Yesterday, I was no longer able to get it to work, on none of my boards. Not the presumably damaged one, not the one it originally worked on, and not the replacement board I have gotten in the meantime. The master just produces three clock pulses on SCL, then nothing. Right now, I'm at a loss on why that is, but I can think of two reasons:
1. There were some special circumstances in my successful test that I fail to replicate now.
2. There is something about my setup the damages the I2C peripheral.

I'll keep investigating. If you, or anyone, is testing this, I'd be very interested in the results.

hannobraun:
Okay, turns out it was a wiring issue in my setup (don't ask about details, unless you're willing to embarrass me in public).

Merging, since @david-sawatzke already approved, and the only thing I did since then was rebase without changes.

PR #255, opened 2020-07-15
Minor SPI clean-ups

hannobraun:


hannobraun:
Upon further research, it looks like a stall is just a temporary condition that happens if software doesn't read or write in time (if I understand the documentation correctly). So it's not something that should return an error, if detected, as in fact the condition will go away if we just proceed with the read/write.

This means that the code in this PR should be correct as-is, and I won't be following up with a pull request adding error handling.

hannobraun:
Pushed another minor clean-up.

PR #258, opened 2020-07-17
Add methods to select and deselect SPI slaves

hannobraun:
Based on #256. I recommend reviewing/merging that one first.

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -57,11 +57,13 @@
 
 mod clock;
 mod instances;
+mod interrupts;
 mod peripheral;
 
 pub use self::{
     clock::{Clock, ClockSource},
     instances::Instance,
+    interrupts::Interrupts,
     peripheral::SPI,
 };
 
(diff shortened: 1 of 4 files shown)


hannobraun:
Hold on, this is not quite right. It turns out that you can't deselect a slave using the method provided here. You do so by setting EOT (end of transfer) in TXCTL or TXDATCTL.

This might mean that we can't provide this feature and implement `spi::FullDuplex` at the same time, as there's no way for `FullDuplex::send` to know that it's writing the last byte. And setting EOT in a separate method will not take effect until we write another word, and would be of no use to a driver anyway.

Probably best to close this PR and handle slave select via GPIO (which I think drivers in the wild already do).

hannobraun:
I'm closing this pull request, as I suggested in my previous comment. It's partially wrong and largely unnecessary in master mode. I've incorporated the parts of it required for slave mode into my WIP branch.

PR #259, opened 2020-07-21
Add support for SPI slave mode

hannobraun:


src/spi/peripheral.rs
```diff
@@ -108,14 +123,17 @@ where
                     w.cpha().set_bit();
                 }
             }
+            match master {
```
david-sawatzke, line 150: Certainly not a needed & not a blocker, but I'd prefer using an enum instead of a bool here, an obvious candidate would be [pac::spi0::cfg::MASTER_A](https://docs.rs/lpc845-pac/0.3.0/lpc845_pac/spi0/cfg/enum.MASTER_A.html)

hannobraun: Good point, no idea why I didn't do that in the first place.

Pushed a commit to address this.

PR #260, opened 2020-07-22
Clean up DMA code

hannobraun:
This PR is a meandering journey through DMA, cleaning up what needs cleaning, fixing some things, but not adding any new features. The most significant change is that a panic on LCP845 was fixed.


hannobraun:
Force-pushed to remove a few commits, make some other slight modifications, and to sneak in another minor commit.

The peripheral-to-memory transfers I'm currently working on seem to be taking a slightly different direction than expected, and some of the changes made here turned out to be counterproductive. @david-sawatzke, I figured making you review these changes, then making you review the changes undoing them in the next PR, would be inconsiderate, so I chose to clean up this pull request instead.

david-sawatzke:
(Also not a blocker and would necessitate some api changes, but the limit of 1024 should probably be enforced)

hannobraun:
> (Also not a blocker and would necessitate some api changes, but the limit of 1024 should probably be enforced)

Thanks for the review!

I agree, but I'm going to merge now and address this in #261, to save myself the rebase.

PR #261, opened 2020-07-23
Add support for DMA-based reading from USART

### LINKED ISSUE #124: DMA: Support peripheral-to-memory transfers
Transferring from a peripheral to memory is currently not supported.

hannobraun:
Based on #260. I recommend reviewing/merging that one first.

Close #124.

### DIFF
--- a/src/dma/buffer.rs
+++ b/src/dma/buffer.rs
@@ -1,3 +1,5 @@
+use crate::pac::dma0::channel::xfercfg::SRCINC_A;
+
 use super::Source;
 
 impl crate::private::Sealed for &'static [u8] {}
@@ -5,6 +7,10 @@ impl crate::private::Sealed for &'static [u8] {}
 impl Source for &'static [u8] {
     fn is_empty(&self) -> bool {
         self.len() == 0
+    }
+
+    fn increment(&self) -> SRCINC_A {
+        SRCINC_A::WIDTH_X_1
     }
 
     fn transfer_count(&self) -> usize {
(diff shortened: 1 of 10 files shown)


hannobraun:
Rebased, ready for review.

@david-sawatzke: To address your comment on #260, I've pushed a commit that panics, if the buffer is too long. Is that what you had in mind? I don't think a `Result` makes sense, as this is going to be a programmer error that is not recoverable.

src/dma/buffer.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    fn increment(&self) -> SRCINC_A {
+        SRCINC_A::WIDTH_X_1
+    }
+
+    fn transfer_count(&self) -> Option<u16> {
```
david-sawatzke, line 23: This isn't an issue with the internal usage, but as this trait is `pub`, a note about this method only being "valid" when it isn't empty might be nice (same for the `Dest` trait).

This would also make the comment in `start_transfer` clearer.

(Not a blocker, since it's not intended for usage by the end user)

hannobraun: Good point. I pushed a commit improving the documentation of `Source`/`Dest` in general, but when it came to documenting that aspect, it just seemed so finicky. I instead changed the implementation to never underflow, at a slight performance cost.

This could be optimized by improving the design of the traits, but I don't want to get into that right now. I think it's fine to leave that for future work, whenever the need arises.

Let me know what you think.

PR #263, opened 2020-07-28
Clean up and extend DMA transfer API

hannobraun:
Based on #262. I recommend reviewing/merging that one first.


hannobraun:
Rebased, ready for review.

PR #264, opened 2020-07-29
Add DMA support for I2C master mode

hannobraun:
Based on #263. I recommend reviewing/merging that one first.

PR #265, opened 2020-07-30
Add DMA support for SPI master mode

hannobraun:
This pull request is based on #263 and #264. I recommend reviewing/merging those first.

### DIFF
--- a/src/spi/dma.rs
+++ b/src/spi/dma.rs
@@ -93,15 +93,13 @@ where
     ) {
         let rx_payload = match self.rx_transfer.wait() {
             Ok(payload) => payload,
-            // can't happen, as error type is `Void`; but need to make the
-            // compiler happy
-            Err((_, payload)) => payload,
+            // can't happen, as error type is `Void`
+            Err(_) => unreachable!(),
         };
         let tx_payload = match self.tx_transfer.wait() {
             Ok(payload) => payload,
-            // can't happen, as error type is `Void`; but need to make the
-            // compiler happy
-            Err((_, payload)) => payload,
+            // can't happen, as error type is `Void`
+            Err(_) => unreachable!(),
         };
 
         (
--- a/src/spi/dma.rs
+++ b/src/spi/dma.rs
@@ -1,0 +1,190 @@
+use core::marker::PhantomData;
+
+use void::Void;
+
+use crate::{
+    dma::{
+        self,
+        transfer::state::{Ready, Started},
+    },
+    init_state::Enabled,
+    pac::dma0::channel::xfercfg::{DSTINC_A, SRCINC_A},
+};
+
+use super::{Instance, Master, SPI};
+
+/// An SPI/DMA transfer
+///
+/// Since the SPI peripheral is capable of sending and receiving at the same
+/// time, using the same buffer, it needs this bespoke `Transfer` struct, which
+/// wraps and manages two `dma::Transfer` structs under the hood.
+pub struct Transfer<State, I: Instance> {
+    spi: SPI<I, Enabled<Master>>,
+    buffer: &'static mut [u8],
+    rx_transfer: dma::Transfer<State, I::RxChannel, Rx<I>, dma::Buffer>,
+    tx_transfer: dma::Transfer<State, I::TxChannel, dma::Buffer, Tx<I>>,
+}
+
+impl<I> Transfer<Ready, I>
+where
+    I: Instance,
+{
+    pub(super) fn new(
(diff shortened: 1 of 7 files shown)


hannobraun:
Rebased, still blocked on #264.

hannobraun:
Rebased, ready for review.

src/spi/dma.rs
```diff
@@ -0,0 +1,190 @@
...
+        SPI<I, Enabled<Master>>,
+        &'static mut [u8],
+        dma::Channel<I::RxChannel, Enabled>,
+        dma::Channel<I::TxChannel, Enabled>,
+    ) {
+        let rx_payload = match self.rx_transfer.wait() {
+            Ok(payload) => payload,
+            // can't happen, as error type is `Void`; but need to make the
+            // compiler happy
+            Err((_, payload)) => payload,
```
david-sawatzke, line 98: Might wanna use `unreachable!()` here
```suggestion
            Err((_, payload)) => unreachable(),
```

src/spi/dma.rs
```diff
@@ -0,0 +1,190 @@
...
+        };
+        let tx_payload = match self.tx_transfer.wait() {
+            Ok(payload) => payload,
+            // can't happen, as error type is `Void`; but need to make the
+            // compiler happy
+            Err((_, payload)) => payload,
```
david-sawatzke, line 104: Also here
```suggestion
            Err((_, payload)) => unreachable(),
```

src/spi/dma.rs
```diff
@@ -0,0 +1,190 @@
...
+        self,
+        transfer::state::{Ready, Started},
+    },
+    init_state::Enabled,
+    pac::dma0::channel::xfercfg::{DSTINC_A, SRCINC_A},
+};
+
+use super::{Instance, Master, SPI};
+
+/// An SPI/DMA transfer
```
david-sawatzke, line 16: Maybe using `transaction` for this would be clearer, since we're not only transferring data in one direction, we're exchanging it in both

src/spi/dma.rs
```diff
@@ -0,0 +1,190 @@
...
+            spi,
+            buffer,
+            rx_transfer,
+            tx_transfer,
+        }
+    }
+
+    /// Start the transfer
+    ///
+    /// Starts both DMA transfers that are part of this SPI transfer.
```
david-sawatzke, line 68: (As an example, in case you agree)
```suggestion
    /// Starts both DMA transfers that are part of this SPI transaction.
```

src/spi/dma.rs
hannobraun: Thank you, I agree this is better. I'll apply this manually later today (your suggestion is missing a `!`).

src/spi/dma.rs
hannobraun: Same as comment above: Will apply manually later.

src/spi/dma.rs
hannobraun: I've been following the naming convention from `embedded-hal`: https://docs.rs/embedded-hal/0.2.4/embedded_hal/blocking/spi/trait.Transfer.html

I'm inclined to keep this as it is, but I don't have a strong opinion. I'll change the name, if you want me to.

src/spi/dma.rs
hannobraun: See comment above.

hannobraun:
> A usage example or something similar would also help (maybe a loopback detection, as a self contained system?).

I do have a usage example over here: https://github.com/braun-embedded/lpc845-test-stand/blob/22de685de7664273169232f20b330954512fc007/test-target/src/main.rs#L621-L650

I agree that it would be better to have a self-contained example using loopback mode in this repository, but I'd like to leave that as a possible future improvement, if at all possible. I'm already very late in my current project, and my focus right now is to finish up the last few tasks so I can take a long vacation. For this reason, I'd like to avoid any additional non-trivial tasks (which this is, as it includes API design work).

src/spi/dma.rs
david-sawatzke: That's a good point, I'm inclined to leave it since the precedent is strong

hannobraun: For what it's worth, I think it would be better, if `embedded-hal` called it "transaction" (and we would follow that). But that's not a battle I'm inclined to fight :-)

hannobraun:
Pushed a commit that adds `unreachable!` in the appropriate places. Rebased everything on latest master (no changes were necessary).

Merging, as everything is resolved now (and my test suite still runs fine).

PR #266, opened 2020-07-30
Add support for changing USART settings

hannobraun:
This is mostly restricted to settings that don't require new modes (like synchronous or RS-485 mode) to be added.


hannobraun:
There were conflicts with #263 that I was unaware of. Rebased.

hannobraun:
> I'm not too happy about the added complexity in the trivial case, but it's not too bad

We can always make improvements later, if you have any ideas. This is the best I could come up with at this time. I don't think we can do much better without losing type safety.

PR #268, opened 2020-08-04
Add missing `Default` implementation for `usart::Interrupts`

hannobraun:


david-sawatzke:
An example/documentation of how this all fits together with the enable_interrupts call may be nice, even though though it's probably more straightforward that you just need to set some fields with the compiled documentation, instead of macros.

hannobraun:
Pushed a commit that adds examples to all the methods that enable/disable USART interrupts. Merging, as the rest was already approved.

PR #269, opened 2020-08-06
Add support for USART hardware flow control

hannobraun:


src/usart/tx.rs
```diff
@@ -278,6 +278,38 @@ where
...
+{
+    /// Disable throttling via CTS signal
+    ///
+    /// Configure the transmitter to ignore the CTS signal. Returns the SWM
+    /// function for the CTS signal, so it can be reused to enable CTS
+    /// throttling again, or for something else.
+    pub fn disable_cts_throttling(
+        self,
+    ) -> (Tx<I, Enabled<W>, NoThrottle>, Function) {
+        interrupt::free(|_| {
```
david-sawatzke, line 294: Not sure if you're planning to change that later, but this currently isn't needed, since we're only changing this in the Usart enable function and in `Tx` stuff. I also prefer if users have the option to supply the `cs` separately, in case they know better (e.g. when using rtic) & not disabling interrupts, but there are definitely pros and cons to that approach as well.

hannobraun: Even if we could do without a critical section here, I don't think we should, unless we're very sure the API is no longer going to change much. It's way too easy to make a change elsewhere and invalidate the assumptions made here.

Having the user supply the critical section token is an interesting idea. I'll look into it later. I'm a bit concerned how the move required by this API is going to interact with an `interrupt::free` closure at the call site, but I think it shouldn't be a problem.

hannobraun:
Rebased on top of master. No changes were necessary.

@david-sawatzke: I thought a bit about your suggestion to have the user pass the critical section token. I think that's definitely better, but it's one more thing that complicates the API and needs to be documented. I then considered to accept the token in an `Option`, to ease the burden on the user, but couldn't quite decide whether that's a good pattern or over-engineering.

In the end, I took the easy way out and decided to leave it as is (you already approved it, after all). I would welcome a future improvement here, but think for now it is fine. The critical section is really short, so the impact is minimal anyway.

PR #282, opened 2020-09-27
Take unshifted I2C addresses.

jacobbramley:
This brings `lpc8xx-hal::i2c` in line with (my understanding of) the [`embedded_hal::i2c::Write`](https://docs.rs/embedded-hal/0.2.4/embedded_hal/blocking/i2c/index.html) trait, which seems to require plain seven-bit addresses.

I tested this with the [`ssd1306` (OLED display driver) crate](https://crates.io/crates/ssd1306), which defaults to the address used by the hardware. With this patch, the two crates work together without issue.

I also updated the examples to keep the actual address unchanged, but I don't have any of the devices they use. Similarly, I don't have a convenient way to test the slave mode.

### DIFF
--- a/examples/i2c_eeprom.rs
+++ b/examples/i2c_eeprom.rs
@@ -93,7 +93,7 @@ fn main() -> ! {
 
     // Address of the eeprom
     // ADJUST THIS
-    let address = 0b1010_0000;
+    let address = 0b101_0000;
 
     serial
         .bwrite_all(b"Writing data...\n")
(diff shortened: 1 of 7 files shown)


david-sawatzke:
So, my reading of the doc trait is the same as yours and others also take 7 bit numbers:
- https://docs.rs/stm32f4xx-hal/0.8.3/src/stm32f4xx_hal/i2c.rs.html#725
- https://atsamd-rs.github.io/atsamd/atsamd21g18a/src/atsamd21_hal/sercom/i2c.rs.html#117

david-sawatzke (approved):

hannobraun (approved):
Thank you, @jacobbramley!

I agree that this is a better way to handle this. I will test slave mode later (using [`lpc845-test-stand`](https://github.com/braun-embedded/lpc845-test-stand)) and will make corrections, if necessary.

PR #283, opened 2020-09-28
Fix handling of I2C slave address

hannobraun:
Follow-up to #282 by @jacobbramley. Turns out we missed one place that had to be modified.


david-sawatzke (approved):
Argh, I went over the code again to look for other locations that could also need it but missed this one. Looks good!

hannobraun:
I missed it too. At least it's confirmed to work now.

PR #287, opened 2020-10-12
Add partial support for embedded-hal 1.0.0-alpha.1

hannobraun:
I've decided to start with the traits that I need right now (GPIO and delay), rather than trying to do everything, which I neither have the time to do nor to test. I figure we can extend support as needed.

Contains some other random commits in addition to the embedded-hal stuff. Those were things I happened upon while working on this, and I figure they're minor enough to not need their own pull request.

### DIFF
--- a/src/delay.rs
+++ b/src/delay.rs
@@ -80,8 +80,6 @@ impl DelayMsAlpha<u32> for Delay {
     type Error = Void;
 
     /// Pauses execution for `ms` milliseconds
-    // At 30 MHz (the maximum frequency), calling delay_us with ms * 1_000 directly overflows at 0x418937 (over the max u16 value)
-    // So we implement a separate, higher level, delay loop
     fn try_delay_ms(&mut self, ms: u32) -> Result<(), Self::Error> {
         Ok(self.delay_ms(ms))
     }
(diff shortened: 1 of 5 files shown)


src/delay.rs
```diff
@@ -81,13 +96,35 @@ impl DelayMs<u16> for Delay {
...
     fn delay_ms(&mut self, ms: u8) {
         self.delay_ms(ms as u16);
     }
 }
 
+impl DelayMsAlpha<u8> for Delay {
+    type Error = Void;
+
+    /// Pauses execution for `ms` milliseconds
+    // At 30 MHz (the maximum frequency), calling delay_us with ms * 1_000 directly overflows at 0x418937 (over the max u16 value)
```
david-sawatzke, line 121: Same here

hannobraun:
Thanks for the review! It baffles me that I managed to find and correct subtle changes in some comments, but then somehow copy-pasted the large inappropriate one without much thought. I'll push an update later.

> Not sure how I feel about having editor specific config, but VSCode makes the most sense (unfortunately, emacs4ever!).

As long as it's not some crazy file format that requires regular maintenance, this kind of configuration is very helpful to users of the editor in question, while having little to no impact on users of other editors. I see little reason to reject this kind of thing, especially from regular contributors. I'd be happy to merge Emacs configuration, if that's helpful to you.

hannobraun:
Rebased (no changes required); added a commit that shortens the inappropriate comments.

david-sawatzke (approved):
Looks great now! I have nothing emacs specific, at least nothing that would make sense.

PR #289, opened 2020-10-16
Add embedded-hal alpha and embedded-time support to MRT

hannobraun:


david-sawatzke (approved):

hannobraun:
CI fails because embedded-time doesn't compile on nightly. See https://github.com/FluenTech/embedded-time/pull/78 and https://github.com/rust-lang/rust/issues/77638. Since a fix has already been merged in embedded-time, I assume it will make its way to crates.io sooner or later, and we won't have issues here.

Merging, as the PR is already approved.

PR #290, opened 2020-10-22
Update and pin embedded-hal alpha version

hannobraun:
It has recently come to my attention that Cargo treats different alpha
versions as semver-compatible. This is bad, as those could contain
breaking changes, and a simple `cargo update` by a user of LPC8xx HAL
could break their build.

Pinning to one specific alpha version should fix that particular
problem.


david-sawatzke (approved):
Great! For future reference: https://github.com/rust-embedded/embedded-hal/issues/177#issuecomment-713525368

PR #291, opened 2020-10-26
Clean up CTIMER; implement `embedded_hal::Pwm`

hannobraun:
Lots of commits in this one, but I believe each one should be relatively easy to review.

As noted in the commit, the implementation of `embedded_hal::Pwm` is _really_ ugly. I'm not sure what else to do though. Once const generics are fully supported, it might might sense to replace the enum with a number and make it all nice, but for now, I don't really know if that mess is worth the improved reliability and runtime overhead.

Happy to hear suggestions.


hannobraun:
Did a bit of force-pushing to fix some CI issues. Nightly still fails for unrelated reasons.

Should be good to go now.

src/ctimer/mod.rs
```diff
@@ -0,0 +1,49 @@
+//! API for the CTimer peripheral
```
david-sawatzke, line 1: Small nit remaining from a15e2a0ec9f8ca917485d0f2403416f6e8c187b7, this should probably also be `CTIMER`

david-sawatzke (approved):
Seems good, but the hal impls really are something else

src/ctimer/mod.rs
hannobraun: Pushed a commit addressing that.

hannobraun:
Thanks for the review!

I've had this idea that with fully-formed const generics, it might be possible to replace the enums with a number and do some type-level validation depending on how many channels have been attached. At this point this is just science fiction, but who knows!

PR #292, opened 2020-10-26
Implement `Pwm` from embedded-hal#246

hannobraun:
This is based on #291. The only new commit is the last one.


hannobraun:
The embedded-hal PR this provides an implementation for has been merged, so this is no longer needed. Closing.

PR #293, opened 2020-11-13
Update embedded-hal dependency to latest version

hannobraun:


hannobraun:
Nightly still fails due to `embedded-time`, but otherwise all is green.

david-sawatzke (approved):

PR #295, opened 2020-11-23
Update dependencies, adapt trybuild tests to latest stable Rust version, fix an example

hannobraun:
Just some general housekeeping to keep things fresh and tidy.


hannobraun:
Pushed another commit that cleans up a minor annoyance that's been bothering me for a while.

Cargo.toml
```diff
@@ -59,7 +59,7 @@ default-features = false
 
 
 [dev-dependencies]
-cortex-m-rtic = "0.5.5"
+cortex-m-rtic = "0.6.0-alpha.0"
```
david-sawatzke, line 62: Since this an alpha version, this should be an exact match, since semver guarantees don't apply

david-sawatzke (approved):
LGTM otherwise

I'm a bit uncomfortable with having alpha versions in our dependencies. We'll have to be careful come release time, so none slip through.

david-sawatzke (changes requested):
Aaargh, forgot to hit the correct button to `Request changes`

Cargo.toml
hannobraun: Thanks! Somehow I had manged to completely forget this again.

hannobraun:
Thanks for the review! Pushed a commit that pins `cortex-m-rtic`.

> I'm a bit uncomfortable with having alpha versions in our dependencies. We'll have to be careful come release time, so none slip through.

I agree in principle, but I don't think it's a problem currently:

- The `embedded-hal` alpha was added in addition to the regular version, so I don't see a drawback. If I'm missing something, we can make it optional.
- `cortex-m-rtic` is a dev-dependency, so not visible to consumers of the crate.

david-sawatzke (approved):

PR #297, opened 2020-12-02
Add DynamicPin

Lotterleben:
This PR adds a `DynamicPin which can change direction at runtime. I've also taken the liberty to refactor some duplicate code into helpers.

Also adds an example that demonstrates how to use these.

The current limitation is that the pin id is still baked into a DynamicPin's type, which makes it difficult to e.g. pass a bunch of these around in a collection to pick and reconfigure them at runtime. I'm working on a solution for this that's probably a breaking change to this PR, but am only ~halfway there.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -261,7 +261,7 @@ where
 
     /// Set pin direction to dynamic (i.e. changeable at runtime)
     ///
-    /// This method is only available while the pin is in input mode.
+    /// This method is only available when the pin is not already in dynamic mode.
     ///
     /// Consumes the pin instance and returns a new instance that is in dynamic
     /// mode, making the methods to change direction as well as read/set levels
@@ -348,7 +348,7 @@ where
 
     /// Set pin direction to dynamic (i.e. changeable at runtime)
     ///
-    /// This method is only available while the pin is in output mode.
+    /// This method is only available when the pin is not already in dynamic mode.
     ///
     /// Consumes the pin instance and returns a new instance that is in dynamic
     /// mode, making the methods to change direction as well as read/set levels
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -67,7 +67,7 @@ use crate::pac::gpio::{
     PIN0 as PIN, SET0 as SET,
 };
 
-use self::direction::Direction;
+use self::direction::{Direction, DynamicPinErr};
 
 /// Interface to the GPIO peripheral
 ///
@@ -225,7 +225,6 @@ where
         // Since all other instances of `GpioPin` are doing the same, there are
         // no race conditions.
         let gpio = unsafe { &*pac::GPIO::ptr() };
-
         let registers = Registers::new(gpio);
         let direction = D::switch::<T>(&registers, arg);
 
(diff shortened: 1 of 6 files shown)


src/gpio.rs
```diff
@@ -260,6 +259,33 @@ where
         }
     }
 
+    /// Set pin direction to dynamic (i.e. changeable at runtime)
+    ///
+    /// This method is only available while the pin is in input mode.
```
hannobraun, line 264: Technically this is correct, but also a bit confusing since an identical method is available in input mode too. Maybe say that this method is available, unless the pin is already in dynamic mode?

It should also be possible to only write this method once, by defining a `direction::NotDynamic` trait and using `where Direction: NotDynamic`. No need to do this now though.

src/gpio.rs
```diff
@@ -320,6 +346,33 @@ where
         }
     }
 
+    /// Set pin direction to dynamic (i.e. changeable at runtime)
+    ///
+    /// This method is only available while the pin is in output mode.
```
hannobraun, line 351: See note on other `into_dynamic` method above.

src/gpio.rs
```diff
@@ -415,6 +468,216 @@ where
...
+        is_high::<T>(&registers)
+    }
+
+    /// Indicates whether the voltage at this pin is currently set to LOW
+    /// This can be used when the pin is in any direction:
+    ///
+    /// If it is currently an Output pin, it indicates whether the pin output is set to LOW
+    /// If it is currently an Input pin, it indicates wether the pin input is LOW
+    ///
+    /// This method is only available, if the pin is in the GPIO state.
```
hannobraun, line 578:
```suggestion
    /// This method is only available, if the pin has been set to dynamic mode.
```

src/gpio.rs
```diff
@@ -415,6 +468,216 @@ where
...
+        set_low::<T>(&registers);
+    }
+
+    /// Indicates whether the voltage at this pin is currently set to HIGH
+    /// This can be used when the pin is in any direction:
+    ///
+    /// If it is currently an Output pin, it indicates whether the pin output is set to HIGH
+    /// If it is currently an Input pin, it indicates wether the pin input is HIGH
+    ///
+    /// This method is only available, if the pin is in the GPIO state.
```
hannobraun, line 561:
```suggestion
    /// This method is only available, if the pin has been set to dynamic mode.
```

hannobraun (changes requested):
Thank you, @Lotterleben! This looks great.

The CI build fails, as the example can't be built. I think it's missing an entry in `Cargo.toml`, which would specify the required features.

I've left some other comments, but the CI thing is the only blocker for me. Everything else is just a note or suggestion.

src/gpio.rs
Lotterleben: Ah, yeah that makes sense. I've corrected the comment, but I'm afraid I don't quite get how your `direction::NotDynamic` design would work: The duplicate `into_dynamic()` functions already are part of a `direction::[Input |  Output]` impl, how can two direction traits apply at the same time?

david-sawatzke: The basic idea is that `direction::NotDynamic` is implemented for `direction::[Input | Output]` (i.e. one implementation for `direction::Input` & another one for `direction::Output`, not both at the same time) and `into_dynamic()` is only implemented with `where Direction: NotDynamic`, not for `direction::[Input | Output]` directly. We've already used this pattern [in the past](https://github.com/lpc-rs/lpc8xx-hal/blob/688aba523b12f34559623252959cc618d6d2bf12/src/gpio.rs#L328), but I think we dropped the `Unknown` state & removed it. Hope that helps

david-sawatzke (approved):

src/gpio.rs
Lotterleben: oh, okay! Thank you for the explanation. If that's okay with you I'd postpone this to a later PR; I've written down a TODO for myself about this though.

hannobraun (approved):

hannobraun:
Everything's ready to go now. Thanks again, @Lotterleben!

PR #298, opened 2020-12-17
Improve embedded-time integration of MRT

hannobraun:


examples/gpio_timer.rs
```diff
@@ -3,7 +3,11 @@
 
 extern crate panic_rtt_target;
 
-use lpc8xx_hal::{cortex_m_rt::entry, gpio::Level, prelude::*, Peripherals};
+use core::convert::TryInto as _;
```
david-sawatzke, line 6: Any reason why it's imported as hidden? I don't see why this shouldn't be in the local namespace. 

If this is just your stylistic preference, that's also fine

examples/gpio_timer.rs
```diff
@@ -38,7 +42,8 @@ fn main() -> ! {
     let mut led = led.into_output_pin(token, Level::Low);
 
     // Start the timer with an interval of 12_000_000 ticks
-    timer.start(12_000_000u32);
+    let ticks: Ticks = 12_000_000u32.try_into().unwrap();
```
david-sawatzke: The `u32` suffix doesn't seem to be required anymore. (also very much a nit)

david-sawatzke: whoops, fixed in the next commit

src/mrt.rs
```diff
@@ -84,9 +84,7 @@ where
     /// The `reload` argument must be smaller than or equal to [`MAX_VALUE`].
     ///
     /// [`MAX_VALUE`]: constant.MAX_VALUE.html
-    pub fn start(&mut self, reload: impl Into<Ticks>) {
-        let reload = reload.into();
-
+    pub fn start(&mut self, reload: Ticks) {
```
david-sawatzke, line 92: I think diverging the inherent impls & the trait impls should be done very carefully & I can also imagine some weird error scenarios depending on which traits are in scope. In this case, it's probably acceptable (since there's nothing to `into` it from), but we have to be very careful.

david-sawatzke (approved):
One very small nitpick, that one can reasonably disagree with. LGTM!

examples/gpio_timer.rs
hannobraun: I always do this when I need a trait because I'm using its methods, but don't refer to it by name. Makes it a bit more obvious why it's imported, and keeps the local namespace clean.

But it's just a small thing, not important at all. Happy to change it or leave it, however you prefer.

src/mrt.rs
hannobraun: Good point, didn't think of that. Keep in mind though, that the current situation is only semi-permanent. At some point we'll probably retire support for embedded-hal 0.2.x, and the new methods have a `try_` prefix.

hannobraun: Oh, and the whole point of the inherent methods is that they are already different: The inherent ones are infallible, while the embedded-hal one's are fallible. So the weird errors depending on traits in scope are already possible.

Overall, I think it's a minor issue, and one that will disappear in time.

hannobraun:
Thanks for the review!

Let me know, if you want me to address any of your nitpicks. I have a slight preference to just merge as-is, but I'm happy to make changes, if you prefer.

david-sawatzke:
If you prefer it this way, that's fine for me as well

PR #299, opened 2020-12-23
mrt-Ticks: add conversion to u32

Lotterleben:
As discussed in https://github.com/braun-embedded/embedded-test-stand/issues/108 .

I didn't make any changes to `Channel::value` and `Channel::reload_value` because to me it seemed like they delivered the right type in the situations they were used in, maybe that's a discussion for 2021 :)

### DIFF
--- a/src/mrt.rs
+++ b/src/mrt.rs
@@ -229,6 +229,13 @@ impl Ticks {
     pub unsafe fn from_u32(value: u32) -> Self {
         Self(value)
     }
+
+    /// Returns the number of ticks of this `Tick` instance.
+    /// This method is provided as a fallback to avoid performance overhead.
+    /// You may also use the `Into` implementations instead, if you can afford it.
+    pub fn to_u32(&self) -> u32 {
+        self.0
+    }
 }
 
 impl TryFrom<u32> for Ticks {
@@ -240,6 +247,14 @@ impl TryFrom<u32> for Ticks {
         }
 
         Ok(Self(value))
+    }
+}
+
+// Note that we don't implement `From` (which would imply `Into` as well) since `From` might fail
+// (`u32` value too big) but `Into` won't (if it fits into `Ticks`, it fits into an `u32`)
+impl Into<u32> for Ticks {
+    fn into(self) -> u32 {
+        self.0
     }
 }
 


david-sawatzke (approved):

src/mrt.rs
```diff
@@ -229,6 +229,13 @@ impl Ticks {
     pub unsafe fn from_u32(value: u32) -> Self {
         Self(value)
     }
+
+    /// Returns the number of ticks of this `Tick` instance.
+    /// This method is provided as a fallback to avoid performance overhead.
+    /// You may also use the `Into` implementations instead, if you can afford it.
```
hannobraun, line 235: Is that true? I don't think there should be any performance difference between this and the `Into` implementation (unless the compiler does something weird). I think the advantage of having the inherent method is for visibility (an `Into` implementation can be overlooked easily).

src/mrt.rs
```diff
@@ -243,6 +250,14 @@ impl TryFrom<u32> for Ticks {
     }
 }
 
+// Note that we don't implement `From` (which would imply `Into` as well) since `From` might fail
+// (`u32` value too big) but `Into` won't (if it fits into `Ticks`, it fits into an `u32`)
+impl Into<u32> for Ticks {
```
hannobraun, line 255: Generally, implementing `From` should be preferred to implementing `Into`, as the core library provides a blanket `Into` impl for all `From`s.

`impl From<Ticks> for u32` should be possible in this case, without the compiler complaining, but I might be wrong.

hannobraun (approved):
Thank you, @Lotterleben!

I'm merging this, as it's such a clear improvement and definitely needed, but I've also left some nitpicky comments. Feel free to address them in a follow-up PR, or not, as you prefer.

src/mrt.rs
Lotterleben: From what I've gathered `From` is not allowed to fail, and from what I gather the conversion will fail if you pass `0_xff_fff_fff` / `u32::MAX` (see https://github.com/lpc-rs/lpc8xx-hal/blob/85923c568c07298e3b2b17bbbe81dd13f7f7837c/src/mrt.rs#L241), which is why I picked this route. Did I miss a better way to resolve this?

hannobraun: The conversion from `Ticks` to `u32` (which is what you've implemented here) is infallible. I'm just suggesting another way to formulate the same conversion:
``` rust
impl From<Ticks> for u32 {
    // ...
}
```

Quoting from the [documentation for `Into`](https://doc.rust-lang.org/std/convert/trait.Into.html):
> One should avoid implementing `Into` and implement `From` instead. Implementing `From` automatically provides one with an implementation of `Into` thanks to the blanket implementation in the standard library.

So implementing `Into` should only be done, if `From` can't be implemented for some reason. I think my suggestion above should work though, because `From<Ticks>` is "local enough" (I know there's a better way to say this, but I'm not familiar with all the Rust lingo).

src/mrt.rs
Lotterleben: To be honest I'm not quite sure what my train of thought was here. happy to open a PR that removes this comment

src/mrt.rs
Lotterleben: Ugh! Yeah you're right– I was thinking about it the wrong way round. Will open a PR together with mitigation for your other comment this evening.

PR #301, opened 2021-01-06
Add test suite using defmt-test

hannobraun:
I'd like to easily run tests on all target platforms after making a change, instead of having to manually run examples with different command-line arguments for each target. This pull request adds testing infrastructure based on defmt-test. The test suite only contains some dummy tests right now, to verify that the infrastructure works.

I plan on adding real tests with some changes I'm working on, but I wanted to submit this separately, in case there's anything that needs to be discussed here.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -30,8 +30,10 @@ if [ "$STABLE_CHECKS" = true ]; then
 fi
 
 function build() {
+    TARGET=$1
+
     echo ""
-    echo "### Building target $1"
+    echo "### Building target $TARGET"
     echo ""
 
     # Only run trybuild on the stable channel. Otherwise changes to compiler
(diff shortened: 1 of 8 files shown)


david-sawatzke (changes requested):
Haven't tested it, but seems mostly ready to merge.

I think this should also be compiled by the CI, so people without all the hardware can at least avoid breaking the test build. If you want to add that in a later PR or have good reason not to do so, feel free to merge it.

hannobraun:
@david-sawatzke Thanks for the review!

> I think this should also be compiled by the CI, so people without all the hardware can at least avoid breaking the test build. If you want to add that in a later PR or have good reason not to do so, feel free to merge it.

Good point! I didn't think about that because we can't run it on CI (with the current infrastructure), but of course we can build it. I'll change that later.

hannobraun:
Included building the test suite in the CI build, as requested. Also rebased on master (no changes were necessary).

david-sawatzke (approved):

PR #302, opened 2021-01-06
Rename Cargo configuration file

hannobraun:
For the longest time I didn't know this was possible, but this is much
nicer. In VS Code, it gives me syntax highlighting and other features
(like shortcuts to comment/uncomment multiple lines) out of the box. I
assume the experience is similar in other editors that support TOML.


david-sawatzke (approved):
Nice find, I'll have to keep in mind for other projects

PR #305, opened 2021-01-18
RFC: more generic Dynamic Pins

Lotterleben:
Hi everyone,
the goal of this PR is more to start a discussion rather than getting it merged as-is (as you can see, there's quite a lot of Doc TODOs in there anyway).
The code in this PR... works (I've been using it [in my test-stand fork](https://github.com/Lotterleben/lpc845-test-stand/tree/generic_dynamic_pin) ; beware– there's a lot of code in dire need of cleanup in there), but I'm not convinced it is the best solution.

The problem it tries to solve is:
Currently, Pin IDs are baked into each pin type. This limits the applicability of Dynamic GPIO Pins: While I can define pins to be changeable at runtime (for example in a Test stand context to re-configure the LPC from test case to test case), I can't gather them in a vector, or define functions like `fn initialize_pin(pin: GpioPin<??, Direction::Dynamic>)`, and handling these `Direction::Dynamic` pins involves verbose match statements and repetitive code, which introduce clutter.

The approach in this PR is to define a new type `DynamicGpioPin<D>`, which moves `port` and `mask` out of the `Token` baked into `GpioPin`s and into the `DynamicGpioPin` struct fields. The downside to this is that the usage differs from the usage of `GpioPin`s, there's potentially a bit more code and helpers like `is_high<T: pins::Trait>` (which I've *just* introduced :D) won't work anymore– though this may be fixable with some macros maybe?

So, my discussion questions would be:
- is this a viable approach? If so, it would make sense to get rid of `GpioPin<T, direction::Dynamic>`, no?
- are there other ways to resolve this which you'd prefer over the sketch in this PR?

### DIFF
--- a/examples/gpio_generic.rs
+++ b/examples/gpio_generic.rs
@@ -1,0 +1,73 @@
+#![no_main]
+#![no_std]
+
+extern crate panic_rtt_target;
+
+use lpc8xx_hal::{
+    cortex_m_rt::entry,
+    delay::Delay,
+    gpio::{direction::Dynamic, GpioPin, Level},
+    pins::{DynamicPinDirection, GenericPin},
+    prelude::*,
+    CorePeripherals, Peripherals,
+};
+
+#[entry]
+fn main() -> ! {
+    rtt_target::rtt_init_print!();
+
+    // Get access to the device's peripherals. Since only one instance of this
+    // struct can exist, the call to `take` returns an `Option<Peripherals>`.
+    // If we tried to call the method a second time, it would return `None`, but
+    // we're only calling it the one time here, so we can safely `unwrap` the
+    // `Option` without causing a panic.
+    let cp = CorePeripherals::take().unwrap();
+    let p = Peripherals::take().unwrap();
+
+    // Initialize the APIs of the peripherals we need.
+    let mut delay = Delay::new(cp.SYST);
+
+    let mut syscon = p.SYSCON.split();
+    let gpio = p.GPIO.enable(&mut syscon.handle);
+
+    // Select pins for all three LEDs
+    let (green_led, green_led_token) = (p.pins.pio1_0, gpio.tokens.pio1_0);
+    let (blue_led, blue_led_token) = (p.pins.pio1_1, gpio.tokens.pio1_1);
+    let (red_led, red_led_token) = (p.pins.pio1_2, gpio.tokens.pio1_2);
+
+    // Generate Generic Dynamic Pins from `Token` + `Pin`s and gather them for further batch
+    // processing.
+    let mut leds: [GpioPin<GenericPin, Dynamic>; 3] = [
+        green_led.into_generic_dynamic_pin(
+            green_led_token,
+            Level::High, // led is off initially
+            DynamicPinDirection::Output,
+        ),
+        blue_led.into_generic_dynamic_pin(
+            blue_led_token,
+            Level::High, // led is off initially
+            DynamicPinDirection::Output,
+        ),
+        red_led.into_generic_dynamic_pin(
+            red_led_token,
+            Level::High, // led is off initially
+            DynamicPinDirection::Output,
+        ),
+    ];
+
+    loop {
+        // Blink all LED colors by looping through the array that holds them
+        // This should make the on-board LED blink like this:
+        // 🟢 🔵 🔴 🟢 🔵 🔴 🟢 🔵 🔴 ...
+        for led in leds.iter_mut() {
+            blink_led(led, &mut delay);
+        }
+    }
+}
+
+/// Turn `led` on for 1000 ms
+fn blink_led(led: &mut GpioPin<GenericPin, Dynamic>, delay: &mut Delay) {
+    led.set_low();
+    delay.delay_ms(1_000_u16);
+    led.set_high();
+}
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -53,11 +53,7 @@ use embedded_hal_alpha::digital::{
 };
 use void::Void;
 
-use crate::{
-    init_state, pac,
-    pins::{self, Token},
-    syscon,
-};
+use crate::{init_state, pac, pins, syscon};
 
 #[cfg(feature = "845")]
 use crate::pac::gpio::{CLR, DIRCLR, DIRSET, NOT, PIN, SET};
@@ -187,7 +183,10 @@ impl GPIO<init_state::Enabled> {
     }
 }
 
-/// A pin used for general purpose I/O (GPIO)
+/// A pin used for general purpose I/O (GPIO).
+///
+/// This struct is a wrapper around the representation of a specific pin `P`– it enables said pin
+/// to be used as a GPIO pin.
 ///
 /// You can get access to an instance of this struct by switching a pin to the
 /// GPIO state, using [`Pin::into_input_pin`] or [`Pin::into_output_pin`].
@@ -206,38 +205,42 @@ impl GPIO<init_state::Enabled> {
 /// [`embedded_hal::digital::v2::OutputPin`]: #impl-OutputPin
 /// [`embedded_hal::digital::v2::StatefulOutputPin`]: #impl-StatefulOutputPin
 /// [`embedded_hal::digital::v2::ToggleableOutputPin`]: #impl-ToggleableOutputPin
-pub struct GpioPin<T, D> {
-    token: pins::Token<T, init_state::Enabled>,
+pub struct GpioPin<P, D> {
+    inner: P, // holds port, id and mask for this specific pin
     _direction: D,
 }
 
-impl<T, D> GpioPin<T, D>
+impl<P, D> GpioPin<P, D>
 where
-    T: pins::Trait,
+    P: pins::Trait,
     D: Direction,
 {
-    pub(crate) fn new(
-        token: Token<T, init_state::Enabled>,
-        arg: D::SwitchArg,
-    ) -> Self {
+    pub(crate) fn new(inner: P, arg: D::SwitchArg) -> Self {
         // This is sound, as we only write to stateless registers, restricting
-        // ourselves to the bit that belongs to the pin represented by `T`.
+        // ourselves to the bit that belongs to the pin represented by `P`.
         // Since all other instances of `GpioPin` are doing the same, there are
         // no race conditions.
         let gpio = unsafe { &*pac::GPIO::ptr() };
         let registers = Registers::new(gpio);
-        let direction = D::switch::<T>(&registers, arg);
+        let direction = D::switch(&registers, arg, &inner);
 
         Self {
-            token,
+            inner,
             _direction: direction,
         }
     }
+
+    /// Get identifying information about this pin in the form of a [`pins::Trait`]
+    ///
+    /// [`pins::Trait`]: ../pins/trait.Trait.html
+    pub fn inner(&self) -> &P {
+        &self.inner
+    }
 }
 
-impl<T> GpioPin<T, direction::Input>
+impl<P> GpioPin<P, direction::Input>
 where
-    T: pins::Trait,
+    P: pins::Trait,
 {
     /// Set pin direction to output
     ///
@@ -245,16 +248,17 @@ where
     ///
     /// Consumes the pin instance and returns a new instance that is in output
     /// mode, making the methods to set the output level available.
-    pub fn into_output(self, initial: Level) -> GpioPin<T, direction::Output> {
+    pub fn into_output(self, initial: Level) -> GpioPin<P, direction::Output> {
         // This is sound, as we only do a stateless write to a bit that no other
         // `GpioPin` instance writes to.
         let gpio = unsafe { &*pac::GPIO::ptr() };
         let registers = Registers::new(gpio);
 
-        let direction = direction::Output::switch::<T>(&registers, initial);
+        let direction =
+            direction::Output::switch(&registers, initial, self.inner());
 
         GpioPin {
-            token: self.token,
+            inner: self.inner,
             _direction: direction,
         }
     }
@@ -270,19 +274,22 @@ where
         self,
         initial_level: Level,
         initial_direction: pins::DynamicPinDirection,
-    ) -> GpioPin<T, direction::Dynamic> {
+    ) -> GpioPin<P, direction::Dynamic> {
         // This is sound, as we only do a stateless write to a bit that no other
         // `GpioPin` instance writes to.
         let gpio = unsafe { &*pac::GPIO::ptr() };
         let registers = Registers::new(gpio);
 
+        // always switch to ensure initial level and direction are set correctly
+        let new_direction = direction::Dynamic::switch(
+            &registers,
+            (initial_level, initial_direction),
+            self.inner(),
+        );
+
         GpioPin {
-            token: self.token,
-            // always switch to ensure initial level and direction are set correctly
-            _direction: direction::Dynamic::switch::<T>(
-                &registers,
-                (initial_level, initial_direction),
-            ),
+            inner: self.inner,
+            _direction: new_direction,
         }
     }
 
@@ -303,7 +310,7 @@ where
         let gpio = unsafe { &*pac::GPIO::ptr() };
         let registers = Registers::new(gpio);
 
-        is_high::<T>(&registers)
+        is_high(&registers, self.inner())
     }
 
     /// Indicates wether the pin input is LOW
@@ -322,9 +329,9 @@ where
     }
 }
 
-impl<T> GpioPin<T, direction::Output>
+impl<P> GpioPin<P, direction::Output>
 where
-    T: pins::Trait,
+    P: pins::Trait,
 {
     /// Set pin direction to input
     ///
@@ -332,16 +339,16 @@ where
     ///
     /// Consumes the pin instance and returns a new instance that is in output
     /// mode, making the methods to set the output level available.
-    pub fn into_input(self) -> GpioPin<T, direction::Input> {
+    pub fn into_input(self) -> GpioPin<P, direction::Input> {
         // This is sound, as we only do a stateless write to a bit that no other
         // `GpioPin` instance writes to.
         let gpio = unsafe { &*pac::GPIO::ptr() };
         let registers = Registers::new(gpio);
 
-        let direction = direction::Input::switch::<T>(&registers, ());
+        let direction = direction::Input::switch(&registers, (), &self.inner);
 
         GpioPin {
-            token: self.token,
+            inner: self.inner,
             _direction: direction,
         }
     }
@@ -357,19 +364,22 @@ where
         self,
         initial_level: Level,
         initial_direction: pins::DynamicPinDirection,
-    ) -> GpioPin<T, direction::Dynamic> {
+    ) -> GpioPin<P, direction::Dynamic> {
         // This is sound, as we only do a stateless write to a bit that no other
         // `GpioPin` instance writes to.
         let gpio = unsafe { &*pac::GPIO::ptr() };
         let registers = Registers::new(gpio);
 
+        // always switch to ensure initial level and direction are set correctly
+        let new_direction = direction::Dynamic::switch(
+            &registers,
+            (initial_level, initial_direction),
+            &self.inner,
+        );
+
         GpioPin {
-            token: self.token,
-            // always switch to ensure initial level and direction are set correctly
-            _direction: direction::Dynamic::switch::<T>(
-                &registers,
-                (initial_level, initial_direction),
-            ),
+            inner: self.inner,
+            _direction: new_direction,
         }
     }
 
@@ -390,7 +400,7 @@ where
         let gpio = unsafe { &*pac::GPIO::ptr() };
         let registers = Registers::new(gpio);
 
-        set_high::<T>(&registers);
+        set_high(&registers, self.inner());
     }
 
     /// Set the pin output to LOW
@@ -410,7 +420,7 @@ where
         let gpio = unsafe { &*pac::GPIO::ptr() };
         let registers = Registers::new(gpio);
 
-        set_low::<T>(&registers);
+        set_low(&registers, self.inner());
     }
 
     /// Indicates whether the pin output is currently set to HIGH
@@ -429,7 +439,7 @@ where
         let gpio = unsafe { &*pac::GPIO::ptr() };
         let registers = Registers::new(gpio);
 
-        is_high::<T>(&registers)
+        is_high(&registers, self.inner())
     }
 
     /// Indicates whether the pin output is currently set to LOW
@@ -464,13 +474,14 @@ where
         let gpio = unsafe { &*pac::GPIO::ptr() };
         let registers = Registers::new(gpio);
 
-        registers.not[T::PORT].write(|w| unsafe { w.notp().bits(T::MASK) });
+        registers.not[self.inner().port()]
+            .write(|w| unsafe { w.notp().bits(self.inner().mask()) });
     }
 }
 
-impl<T> GpioPin<T, direction::Dynamic>
+impl<P> GpioPin<P, direction::Dynamic>
 where
-    T: pins::Trait,
+    P: pins::Trait,
 {
     /// Tell us whether this pin's direction is currently set to Output.
     pub fn direction_is_output(&self) -> bool {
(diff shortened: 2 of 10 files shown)


hannobraun (commented):
Thank you, @Lotterleben!

I'm wondering if we even need `DynamicGpioPin`. Here's how I would approach this:

- Change constants in `pins::Trait` to (non-const) methods that take `&self`. If the pin is statically known, the compiler still has everything it needs to optimize this out, so hopefully this won't degrade performance.
- Since the types that implement `pins::Trait` are embedded in `pins::Token`, we either need to make that field `pub(crate)`, or add a `pub fn pin(&self) -> &T` to `Token` (the second one is probably better; provides more flexibility to users and I can't think of a disadvantage off the top of my head).
- Add struct `DynamicPin` somewhere (probably in `pins`, rather than `gpio`), which implements `pins::Trait`, but returns dynamic data, not constants. Make sure it can't be constructed from outside the crate, as to not mess with the uniqueness of pins.
- Add `Pin::into_dynamic`, which transforms a `Pin<T, State>` into a `Pin<DynamicPin, State>`. The `DynamicPin`'s fields are initialized from the values that the `<T as pins::Trait>` methods return.

Unless I'm missing something, this should make `DynamicPin` a drop-in replacement for the other `pins::Trait` implementations and not require any new APIs or duplicated code. What do you think?

Lotterleben:
(just a short heartbeat, I haven't abandoned this PR, it's just taking me a while to test your suggestions between other work)

hannobraun:
Thanks for the update. Let me know, if there's anything I can help with.

Lotterleben:
A little refactoring side note: having `pins::Trait` and `Token`s both be `<T>` gets in the way of readability a bit I think– I keep mixing them up when untangling what goes where :D

src/pins/gen.rs
```diff
@@ -52,12 +52,32 @@ macro_rules! pins {
             ///
             /// [`Pin`]: struct.Pin.html
             #[allow(non_camel_case_types)]
-            pub struct $type(());
+            pub struct $type {
+                PORT: usize,
+                ID  : u8,
+                MASK: u32,
+            }
```
hannobraun, line 59: I think this is fine for a first version, but not ideal. We know the correct values of these for `$type` and can just return those values from the methods directly. No need to save them in the struct, which is uncessary overhead.

Also, as a general note: I prefer established Rust naming convention, so `port`, `id`, and `mask`. This crate does have a lot of all-caps names, but they usually refer to names that are the same in the user manual.

src/pins/gen.rs
```diff
@@ -103,7 +123,8 @@ macro_rules! pins {
         /// [`GPIO`] for more information.
         ///
         /// [`GPIO`]: ../gpio/struct.GPIO.html
-        pub struct Token<T, State>(T, PhantomData<State>);
+        // TODO impl getter instead of making T pub
+        pub struct Token<Trait: ?Sized, State>(pub Trait, PhantomData<State>);
```
hannobraun, line 127: Why the `?Sized` bound? Why wouldn't we always know the size of `pins::Trait` implementations? This is also causing the CI failure.

I see no reason not to leave this as it was (besides `pub` or accessor method; of course, I might be missing something).

hannobraun (commented):
Hey @Lotterleben, thanks for your work so far! I've taken a look at the latest diff, and here are some thoughts.

src/pins/gen.rs
Lotterleben: ah right, that makes sense. Will change!

src/pins/gen.rs
Lotterleben: The honest answer is: because rustc suggested it and the resulting error was more concise – but the "before" error afaik says the same:

```
error[E0277]: the size for values of type `(dyn pins::traits::Trait + 'static)` cannot be known at compilation time
   --> src/gpio.rs:386:16
    |
386 |         token: Token<dyn pins::Trait, init_state::Enabled>,
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
   ::: src/pins/gen.rs:131:26
    |
131 |         pub struct Token<Trait, State>(pub Trait, PhantomData<State>);
    |                          ----- required by this bound in `pins::gen::Token`
    |
    = help: the trait `Sized` is not implemented for `(dyn pins::traits::Trait + 'static)`
help: you could relax the implicit `Sized` bound on `Trait` if it were used through indirection like `&Trait` or `Box<Trait>`
   --> src/pins/gen.rs:131:26
    |
131 |           pub struct Token<Trait, State>(pub Trait, PhantomData<State>);
    |                            ^^^^^             ----- ...if indirection was used here: `Box<Trait>`
    |                            |
    |                            this could be changed to `Trait: ?Sized`...
...
169 | / pins!(
170 | |     pio0_0 , PIO0_0 , 0, 0x00, state::Unused;
171 | |     pio0_1 , PIO0_1 , 0, 0x01, state::Unused;
172 | |     pio0_2 , PIO0_2 , 0, 0x02, state::Swm<((),), ()>;
...   |
223 | |     pio1_21, PIO1_21, 1, 0x15, state::Unused;
224 | | );
    | |__- in this macro invocation
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

From my understanding, we can't know the size of all possible `pins::Trait` implementations as soon as the Trait contains methods instead of only consts. Our concrete implementation in `pins!` is known, but that doesn't make it hold true for the rest of the universe? Especially not if we re-add a `struct DynamicPin` that returns dynamic values for `pins::Trait`?

hannobraun: We don't need to know the size of all possible `pins::Trait` implementations! So far, the API handles this using type parameters:

``` rust
fn use_pin<Pin>(pin: &mut Pin) {}
```
`Pin` is just a placeholder. Code for this function will only be generated, if you actually call it, and then the compiler will know the concrete type (and therefore its size). Yes, the calling function could itself have a generic `Pin`, but then someone is going to call that function (or some function up the chain) with a concrete type, or no code is ever going to be generated.

That we don't know the size of all possible `pins::Trait` implementations only becomes a problem once you make it one. Which you did with this bit (quoting your error message): `Token<dyn pins::Trait, init_state::Enabled>`. `dyn Trait` means "any implementation of trait", which the compiler can't know the size for. Since type parameters carry an implicit `Sized` bound, the compiler rightly complains that you can't use that here.

But the compiler's suggestion is misleading, because (as far as I can tell), there never was a good reason to use `dyn Trait` in the first place. Also, please note the following bit from the error message (emphasis mine):
> you could relax the implicit `Sized` bound on `Trait` **if it were used through indirection like `&Trait` or `Box<Trait>`**

Telling the compiler "this type doesn't need to have a known size" doesn't help, if the type is used for a struct field. There's no way to generate machine code from that. So it rightly tells you, that you also need indirection here, via `&Trait` or `Box<Trait>`. A pointer has a defined size, so we can put it in a struct. That the pointer points to something with an unknown size doesn't matter.

But in our case, I think this indirection is unnecessary. Using type parameters in combination with trait bounds everywhere, as before, should work fine.

Lotterleben:
Hi, another set of follow-up questions regarding these steps:

> * Add struct `DynamicPin` somewhere (probably in `pins`, rather than `gpio`), which implements `pins::Trait`, but returns dynamic data, not constants. Make sure it can't be constructed from outside the crate, as to not mess with the uniqueness of pins.
> * Add `Pin::into_dynamic`, which transforms a `Pin<T, State>` into a `Pin<DynamicPin, State>`. The `DynamicPin`'s fields are initialized from the values that the `<T as pins::Trait>` methods return.
> 
> Unless I'm missing something, this should make `DynamicPin` a drop-in replacement for the other `pins::Trait` implementations and not require any new APIs or duplicated code. What do you think?

- Why the `State` in `Pin<DynamicPin, State>` ? From my perspective, all `DynamicPin`s should be active GPIO pins for now, I'm not sure if I am able to fully assess the consequences of e.g. swm pins being dynamic
- From my perspective, we'd still need to duplicate a lot of `GpioPin` impl code since our `DynamicPin` needs to be able to do similar tings, which clashes with your "no duplicated code" assessment. If we were able to construct something like 

```rust
// in pin impl
    pub fn into_dynamic_pin(
        self,
        token: Token<T, init_state::Enabled>,
        level: Level,
        direction: DynamicPinDirection,
    ) -> GpioPin<T, direction::Dynamic> {
    //    ^^^^^^^^^ Note: we're not returning a `Pin` as suggested but rather a `GpioPin`
        generic_token = // somehow consume `token` and reconstruct it with a `DynamicPin` instance as T
        GpioPin::new(generic_token, (level, direction))
    }
```
we would be able to prevent that, though?

(Side note: a `DynamicPin` struct is easy to confuse with `diection::Dynamic` so I'd rather give it a different name, maybe `GenericPin`?)

hannobraun:
Good observation, I agree with your assessment. `Pin<DynamicPin, State>` was a brain fart :-)


> (Side note: a `DynamicPin` struct is easy to confuse with `diection::Dynamic` so I'd rather give it a different name, maybe `GenericPin`?)

All names in my comments should be taken as just a suggestion. I agree that `GenericPin` seems clearer.

Lotterleben:
aaaand `examples/gpio_generic.rs` blinks! ✨ 
https://user-images.githubusercontent.com/581552/107791636-77d41400-6d54-11eb-8db3-44b531ec15ac.mov

Drop-in replacement works now, doc TODOs will follow asap.

hannobraun:
Nice, good job! I had a quick glance and it looks good. Let me know when this is ready for review.

Lotterleben:
> Nice, good job! I had a quick glance and it looks good. Let me know when this is ready for review.

ready :) (needs to be squashed before merging of course)
I've felft in the TODOs to change `port` to `u8` because imo this shouldn't be sneaked in through a PR like this but it's nice to have a reminder– I'd open a follow-up PR right away if you're OK with this approach.

src/pins/gen.rs
```diff
@@ -52,12 +52,21 @@ macro_rules! pins {
             ///
             /// [`Pin`]: struct.Pin.html
             #[allow(non_camel_case_types)]
+            #[derive(Clone, Copy)]
```
hannobraun, line 55: Nope! This means any pin type can be copied around without restrictions. Not sure what the actual consequence of that is, since the GPIO case at least still requires a token, but the API is certainly written with the assumption that pin types are unique. This is bound to break something, at the very least in the SWM API.

src/pins/gen.rs
```diff
@@ -79,7 +88,7 @@ macro_rules! pins {
             pub(crate) fn new() -> Self {
                 Self {
                     $(
-                        $field: Token($type(()), PhantomData),
+                        $field: Token(&$type(()), PhantomData),
```
hannobraun, line 91: See comment on `Token` below. If my suggestion works out, we can switch to this:

```suggestion
                        $field: Token(PhantomData, PhantomData),
```

src/pins/gen.rs
```diff
@@ -97,13 +106,26 @@ macro_rules! pins {
             }
         }
 
-        /// A token representing a pin
+        /// A token representing a pin.
+        /// Tokens are unique for each pin, and ensure that no pins are re-used by accident.
+        /// They also hold the `pin::Trait` information for their associated pin.
```
hannobraun, line 111: This isn't quite correct. Pin types not being `Copy`/`Clone` and having private constructors ensure that pins are unique and not re-used by accident. `Token` is a somewhat convoluted way to ensure the user doesn't disable the GPIO peripheral after having initialized GPIO pins. This is documented in `GPIO`, I think.

```suggestion
```

src/pins/gen.rs
```diff
@@ -97,13 +106,26 @@ macro_rules! pins {
...
+        impl<Trait, State> Token<Trait, State>
+        {
+            /// Get identifying information about the pin representred by this token
+            /// in the form of a [`pins::Trait`]
+            ///
+            /// [`pins::Trait`]: trait.Trait.html
+            pub fn pin_info(&self) -> &Trait{
+                self.0
+            }
+        }
```
hannobraun, line 128: Since we now throw away the token when initializing a GPIO pin, I don't think we need any of this any more.

```suggestion
        pub struct Token<Pin, State>(PhantomData<Pin>, PhantomData<State>);
```

src/pins/pin.rs
```diff
@@ -175,6 +175,20 @@ pub struct Pin<T: Trait, S: State> {
...
 
+/// A drop-in substitute for the `PIO0_0`, `PIO0_1`... etc. structs which are generated at compile
+/// time and used to bind each pin to a [`pins::Token`].
+/// `GenericPin`s are *only* intended for use cases which are made impossible by the default
+/// [`pins::Trait`] having pin information compiled into its type, e.g. if you'd like to store
+/// several pin instances in a collection, or modify them in batches.
+///
+/// [`pins::Token`]: struct.Token.html
+/// [`pins::Trait`]: trait.Trait.html
+#[derive(Clone, Copy)]
```
hannobraun, line 186: This must not be `Clone` or `Copy`, or the uniqueness guarantee no longer holds!

```suggestion
```

src/pins/pin.rs
```diff
@@ -187,7 +201,7 @@ pub enum DynamicPinDirection {
 
 impl<T> Pin<T, state::Unused>
 where
-    T: Trait,
+    T: Copy + Trait,
```
hannobraun, line 204:
```suggestion
    T: Trait,
```

src/pins/pin.rs
```diff
@@ -238,7 +252,7 @@ where
         self,
         token: Token<T, init_state::Enabled>,
     ) -> GpioPin<T, direction::Input> {
-        GpioPin::new(token, ())
+        GpioPin::new(*token.pin_info(), ())
```
hannobraun, line 255: With my suggested change above, we can no longer get this from `Token`, but I don't think we need to:
```suggestion
        GpioPin::new(self.ty, ())
```

src/pins/pin.rs
```diff
@@ -293,7 +307,7 @@ where
         token: Token<T, init_state::Enabled>,
         initial: Level,
     ) -> GpioPin<T, direction::Output> {
-        GpioPin::new(token, initial)
+        GpioPin::new(*token.pin_info(), initial)
```
hannobraun, line 310:
```suggestion
        GpioPin::new(self.ty, initial)
```

src/pins/pin.rs
```diff
@@ -362,7 +376,7 @@ where
         level: Level,
         direction: DynamicPinDirection,
     ) -> GpioPin<T, direction::Dynamic> {
-        GpioPin::new(token, (level, direction))
+        GpioPin::new(*token.pin_info(), (level, direction))
```
hannobraun, line 379:
```suggestion
        GpioPin::new(self.ty, (level, direction))
```

src/pins/pin.rs
```diff
@@ -399,6 +413,82 @@ where
...
+    pub fn into_generic_dynamic_pin(
+        self,
+        token: Token<T, init_state::Enabled>,
+        level: Level,
+        direction: DynamicPinDirection,
+    ) -> GpioPin<GenericPin, direction::Dynamic> {
+        let port = token.pin_info().port();
+        let id = token.pin_info().id();
+
+        GpioPin::new(GenericPin::new(port, id), (level, direction))
```
hannobraun, line 490:
```suggestion
        GpioPin::new(GenericPin::new(self.ty.port(), self.ty.id()), (level, direction))
```

hannobraun (changes requested):
This looks good, mostly. Thanks again, @Lotterleben! A follow-up commit for changing `port` to `u8` sounds good.

I've left a bunch of comments. Most of them are just notes about details (some optional), but I also found some more fundamental problems regarding the uniqueness of pins. Fortunately I think that those can be addressed pretty easily, and I've left suggested changes to do so.

I didn't try to compile any of my suggestions though, so if they don't work out, let me know. Then we might have a bigger problem :smiley: 

Besides that, some overall notes:
- As you note, there are way too many commits. Ideally, we'd have a series of small and self-contained commits (renaming `T` to `P`, transitioning to throwing away the token, deriving `Copy` for `Level`, etc.), each with a concise explanation of the motivation, unless that is self-evident. But at this point, this is going to be more than a little bit of work, so squashing into a single commit is acceptable.
- The build fails because the example doesn't specify the features it needs to be built with in `Cargo.toml`. There are other LPC845-only examples there whose entry can just be adapted.

@david-sawatzke, let me know if you want to take a look, or if I should just merge once the comments are addressed.

src/gpio.rs
```diff
@@ -206,55 +202,63 @@ impl GPIO<init_state::Enabled> {
 /// [`embedded_hal::digital::v2::OutputPin`]: #impl-OutputPin
 /// [`embedded_hal::digital::v2::StatefulOutputPin`]: #impl-StatefulOutputPin
 /// [`embedded_hal::digital::v2::ToggleableOutputPin`]: #impl-ToggleableOutputPin
-pub struct GpioPin<T, D> {
-    token: pins::Token<T, init_state::Enabled>,
+pub struct GpioPin<P: 'static, D> {
+    pin_info: P, // reminder: holds port, id and mask for this specific pin
```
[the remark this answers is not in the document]
hannobraun: Oh, I notice there's another `'static` here that can be removed, if everything works out as I intend.

src/gpio.rs
```diff
@@ -206,55 +202,63 @@ impl GPIO<init_state::Enabled> {
...
-            token,
+            pin_info,
             _direction: direction,
         }
     }
+
+    /// Get identifying information about this pin in the form of a [`pins::Trait`]
+    ///
+    /// [`pins::Trait`]: ../pins/trait.Trait.html
+    pub fn pin_info(&self) -> &P {
```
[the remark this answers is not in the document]
Lotterleben: I'm not super fond of the name either, so open to renamings. That mental model makes a lot of sense and untangles a lot of confusion I went through when trying to wrap my head around the code– I'll mention that in the `GpioPin` docs and rename `pin_info` to `pin` (or `pin_raw`?)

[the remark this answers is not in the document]
hannobraun: Documenting that mental model sounds great!

I think I'd call the field/method `inner`, with `pin` being my second choice. But those are weak preferences. Do as you wish!

src/pins/gen.rs
Lotterleben: Aaaah I was worried about that. The reason I added this was because I couldn't for the life of me get the `pin_info()` data out of `token` otherwise, but with your suggestion to just use `Pin`'s `ty`, which holds the same information, this should be resolved. Neat!

src/pins/gen.rs
Lotterleben: 🤯

src/pins/gen.rs
hannobraun: I think having the pin information in the token was a stupid idea in the first place (I'm allowed to say that, I implemented it :smile:). It's confusing, and there never was any need for it. It just should have been a `PhantomData` from the start.

Glad that's going to get resolved with this PR!

examples/gpio_generic.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    // NOTE: this example is currently only available to the lpc845
+    #[cfg(feature = "845")]
+    let gpio = {
+        let mut syscon = p.SYSCON.split();
+        p.GPIO.enable(&mut syscon.handle)
+    };
+
+    // Select pins for all three LEDs
+    #[cfg(feature = "845")]
```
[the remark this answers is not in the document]
Lotterleben: oh duh! done :)

src/gpio.rs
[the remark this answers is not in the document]
Lotterleben: `inner` it is. done! :)

Lotterleben:
Alright all resolved, checks are green– is this PR ready to squash?

examples/gpio_generic.rs
```diff
@@ -0,0 +1,75 @@
...
+    let cp = CorePeripherals::take().unwrap();
+    let p = Peripherals::take().unwrap();
+
+    // Initialize the APIs of the peripherals we need.
+    let mut delay = Delay::new(cp.SYST);
+
+    let gpio = {
+        let mut syscon = p.SYSCON.split();
+        p.GPIO.enable(&mut syscon.handle)
+    };
```
hannobraun, line 33: Nitpick: The block is unnecessary and drops `syscon`, which anyone who wants to extend this example into something less trivial is going to need. This is probably a copy-past leftover from other GPIO examples who use this technique to paper over the differences between LPC82x and LPC845.

```suggestion
    let mut syscon = p.SYSCON.split();
    let gpio = p.GPIO.enable(&mut syscon.handle);
```

This is just a tiny complaint though. It can just as well stay as is.

src/gpio.rs
```diff
@@ -464,13 +474,16 @@ where
         let gpio = unsafe { &*pac::GPIO::ptr() };
         let registers = Registers::new(gpio);
 
-        registers.not[T::PORT].write(|w| unsafe { w.notp().bits(T::MASK) });
+        let port = self.inner().port();
+
+        registers.not[port]
+            .write(|w| unsafe { w.notp().bits(self.inner().mask()) });
```
hannobraun, line 480: Another tiny nitpick: This is somewhat inconsistent with similar code below.

```suggestion
        registers.not[self.inner().port()]
            .write(|w| unsafe { w.notp().bits(self.inner().mask()) });
```

src/pins/pin.rs
```diff
@@ -236,9 +249,11 @@ where
     /// [`GPIO`]: ../gpio/struct.GPIO.html
     pub fn into_input_pin(
         self,
-        token: Token<T, init_state::Enabled>,
+        _token: Token<T, init_state::Enabled>,
     ) -> GpioPin<T, direction::Input> {
-        GpioPin::new(token, ())
+        // note that `_token` is consumed and discarded at this pint because we don't need it
+        // anymore– it has served its purpose of guaranteeing pin uniqueness
```
hannobraun, line 255:
```suggestion
        // note that `_token` is consumed and discarded at this pint because we don't need it
        // anymore– it has served its purpose of guaranteeing that the user won't accidentally
        // disable the GPIO peripheral while the pin is active
```

src/pins/pin.rs
```diff
@@ -290,10 +305,12 @@ where
     /// [`GPIO`]: ../gpio/struct.GPIO.html
     pub fn into_output_pin(
         self,
-        token: Token<T, init_state::Enabled>,
+        _token: Token<T, init_state::Enabled>,
         initial: Level,
     ) -> GpioPin<T, direction::Output> {
-        GpioPin::new(token, initial)
+        // note that `_token` is consumed and discarded at this pint because we don't need it
+        // anymore– it has served its purpose of guaranteeing pin uniqueness
```
hannobraun, line 312: See suggestion above.

src/pins/pin.rs
```diff
@@ -358,11 +375,13 @@ where
...
     pub fn into_dynamic_pin(
         self,
-        token: Token<T, init_state::Enabled>,
+        _token: Token<T, init_state::Enabled>,
         level: Level,
         direction: DynamicPinDirection,
     ) -> GpioPin<T, direction::Dynamic> {
-        GpioPin::new(token, (level, direction))
+        // note that `_token` is consumed and discarded at this pint because we don't need it
+        // anymore– it has served its purpose of guaranteeing pin uniqueness
```
hannobraun, line 383: See suggestion above.

src/pins/pin.rs
hannobraun, line 255: Alternative suggestion:

```suggestion
        // note that `_token` is consumed and discarded at this pint because we don't need it
        // anymore– it has served its purpose
```

src/pins/pin.rs
```diff
@@ -427,3 +525,24 @@ where
         }
     }
 }
+
+impl GenericPin {
+    /// Creates a new `GenericPin`
+    pub fn new(port: usize, id: u8) -> Self {
```
hannobraun, line 531: Still public, which breaks pin uniqueness!

```suggestion
    pub(super) fn new(port: usize, id: u8) -> Self {
```

I didn't carefully check if `pub(super)` is enough, maybe we need `pub(crate)` here.

hannobraun (changes requested):
Thanks again, @Lotterleben! I went over it again with a fine comb and came up with mostly nitpicks and one remaining issue that breaks pin uniqueness.

Please address the pin uniqueness issue, and probably the comments about the token, since those are somewhat misleading. Do with the rest as you please.

Other than that, this is ready to merge!

src/pins/pin.rs
hannobraun: Alternative suggestion:

```suggestion
        // note that `_token` is consumed and discarded at this pint because we don't need it
        // anymore– it has served its purpose
```

hannobraun (approved):
Looks great now, thanks @Lotterleben!

PR #306, opened 2021-02-17
`pins::Trait`: downsize to `u8`

Lotterleben:
as promised, resize `pins::Trait` to `u8`

Based on https://github.com/lpc-rs/lpc8xx-hal/pull/305 ; so:
- ~relevant commit is the second one: https://github.com/lpc-rs/lpc8xx-hal/commit/8d91838fa454241790c546756866c07f5332b157~
- ~should be rebased & only merged after #305 is merged~

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -474,7 +474,7 @@ where
         let gpio = unsafe { &*pac::GPIO::ptr() };
         let registers = Registers::new(gpio);
 
-        registers.not[self.inner().port()]
+        registers.not[usize::from(self.inner().port())]
             .write(|w| unsafe { w.notp().bits(self.inner().mask()) });
     }
 }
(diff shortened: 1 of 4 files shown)


src/gpio.rs
```diff
@@ -474,7 +474,7 @@ where
         let gpio = unsafe { &*pac::GPIO::ptr() };
         let registers = Registers::new(gpio);
 
-        registers.not[self.inner().port()]
+        registers.not[self.inner().port() as usize]
```
hannobraun, line 477:
```suggestion
        registers.not[usize::from(self.inner().port())]
```
I think this should be possible? If so, I think it's more readable than a cast, as you can never quite know what `as` will do (unless you happen to know which type you're casing from), while `Into`/`From` are infallible.

Not a big thing, especially since we know it can't go wrong in this case. I'm just thinking about some poor reader in the future, getting suspicious about all the `as` (which is `unsafe`s little brother, in my opinion).

hannobraun:
Thank you, @Lotterleben. #305 is merged, so this is ready to rebase.

src/gpio.rs
Lotterleben: haha, ok! I don't have any strong opinions on the matter anyway so– changed to `usize::from()` and rebased ✅
(note for the future: `channel.rs` and `ctimer_blink.rs` have some `as usize` conversions too)

hannobraun: Yeah, I'm sure there's a lot of this in any code base I'm working on. It's the obvious thing to do, so it happens easily. But I think it is a bit of an anti-pattern in most cases.

hannobraun (approved):
Looking good, thank you @Lotterleben!

PR #308, opened 2021-02-25
Improve USART error handling

### LINKED ISSUE #307: usart rx: OVERRUN flag is never cleared
The`OVERRUN` flag is never cleared in https://github.com/lpc-rs/lpc8xx-hal/blob/master/src/usart/rx.rs#L274, making it difficult to recover from a USART overrun.
We've found that modifying `Rx::receive` to clear the flag resolves the issue:
```diff
if stat.overrunint().bit_is_set() {
+    usart.stat.write(|w| w.overrunint().set_bit());
    Err(nb::Error::Other(Error::Overrun))
 }
```

however, there's also the (unused in the hal) function `is_flag_set()` that according to its docs (and from what I understand about its implementation) clears the flag that's being checked as a side effect. Unfortunately, I've been unable to confirm that 

```diff
-if stat.overrunint().bit_is_set() {
+if self.is_flag_set(Flag::OVERRUN) {
    Err(nb::Error::Other(Error::Overrun))
 }
```

resolves the issue as well, but tbh my testing strategies were a bit wonky.

hannobraun:
I've verified that receiving characters via USART still works with this code. But I don't have a setup that would allow me to inject errors, so I can't test whether this has the intended effect. All feedback is appreciated!

Close #307

### DIFF
--- a/src/usart/peripheral.rs
+++ b/src/usart/peripheral.rs
@@ -487,7 +487,7 @@ where
     I: Instance,
     W: Word,
 {
-    type Error = Error;
+    type Error = Error<W>;
 
     /// Reads a single word from the serial interface
     fn read(&mut self) -> nb::Result<W, Self::Error> {
--- a/src/usart/rx.rs
+++ b/src/usart/rx.rs
@@ -253,7 +253,7 @@ where
     I: Instance,
     W: Word,
 {
-    type Error = Error;
+    type Error = Error<W>;
 
     fn read(&mut self) -> nb::Result<W, Self::Error> {
         // Sound, as we're only reading from `stat`, and `rxdatstat` is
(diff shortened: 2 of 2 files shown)


david-sawatzke (approved):
This looks pretty good to me. I also haven't tested this, hopefully @Lotterleben can confirm this works (at least for overflow events), since that parts looks (almost) identical to the one proposed in #307.

Lotterleben:
Trying it out right now, can already confirm that this would be a breaking change for the Test-Adapter which.. could be a problem :S

Lotterleben:
However, looks like it works after I've adjusted the test-stand API (my super scientific method was to run one of my own gpio tests that trigger a lot of messages in a loop: `while true; do cargo test --package lpc845-test-suite --test gpio  assistant_red_led_should_be_toggleable_by_pin_direction; sleep 1; done` and print the `receive()` result in the `usart0` interrupt handler)

hannobraun:
Okay then, sounds to me like this is ready to merge. Further adjustments can be made later, if necessary.

Thanks for reviewing/testing, everyone!

PR #310, opened 2021-02-26
Switch from travis to github ci

david-sawatzke:
Travis ... reduced their free tier some time ago, so no CI runs. Github seems like the obvious pick. This doesn't test on beta anymore, but I don't think we need it.

EDIT: Huh, weird, it doesn't run in this repo. Runs properly in https://github.com/david-sawatzke/lpc8xx-hal/pull/1. No idea what that's about


.github/workflows/ci.yml
```diff
@@ -0,0 +1,31 @@
+on:
+  push:
+    branches: [ master ]
+  pull_request:
```
hannobraun, line 4: I would have expected this:
```suggestion
  pull_request: [ master ]
```

But I think that only limits the builds being run. It shouldn't cause a build to be run that's not run now?

hannobraun:
No idea why it doesn't run. The configuration looks good to me.

Let's just merge and see how it goes.

hannobraun:
It runs for `master` now. Guess we'll have to observe how it behaves for future pull requests.

PR #311, opened 2021-03-05
gpio.rs: add get_level()

Lotterleben:
Hi 👋,
I've found myself creating match statements like these

```
let pinint0_level = match pinint0_pin.is_high() {
    true => pin::Level::High,
    false => pin::Level::Low,
};
```

more often than I'd like, and imo it'd be neat to resolve this at the HAL level.
So this PR adds a native `get_level()` (or `get_set_level()` in the case of `Output` pins) function to make the level directly accessible.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -1030,7 +1030,7 @@ pub mod direction {
     }
 
     /// Error that can be thrown by operations on a Dynamic pin
-    #[derive(Copy, Clone)]
+    #[derive(Copy, Clone, Debug)]
     pub enum DynamicPinErr {
         /// you called a function that is not applicable to the pin's current direction
         WrongDirection,
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -238,7 +238,7 @@ where
     }
 
     /// Indicates wether the voltage at the pin is currently HIGH
-    /// This is not accessible to the user to avoid confusion becauzse `is_high()`
+    /// This is not accessible to the user to avoid confusion because `is_high()`
     /// semantics differ depending on pin direction. It is only used to implement
     /// `is_high()` and `is_set_high()` respectively for the different direction types.
     pub(crate) fn is_high_inner(&self) -> bool {
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -345,24 +345,6 @@ where
     pub fn get_level(&self) -> Level {
         Level::from_pin(&self)
     }
-
-    /// Returns the current voltage level at this pin.
-    ///
-    /// This method is only available, if two conditions are met:
-    /// - The pin is in the GPIO state.
-    /// - The pin direction is set to input.
-    ///
-    /// See [`Pin::into_input_pin`] and [`into_input`]. Unless both of these
-    /// conditions are met, code trying to call this method will not compile.
-    ///
-    /// [`Pin::into_input_pin`]: ../pins/struct.Pin.html#method.into_input_pin
-    /// [`into_input`]: #method.into_input
-    pub fn get_level(&self) -> Level {
-        match self.is_high() {
-            true => Level::High,
-            false => Level::Low,
-        }
-    }
 }
 
 impl<P> GpioPin<P, direction::Output>
@@ -641,22 +623,6 @@ where
     /// Unless this condition is met, code trying to call this method will not compile.
     pub fn get_level(&self) -> Level {
         Level::from_pin(&self)
-    }
-
-    /// Returns the current voltage level at this pin.
-    /// This can be used when the pin is in any direction:
-    ///
-    /// If it is currently an Output pin, it indicates to which level the pin is set
-    /// If it is currently an Input pin, it indicates the level currently present at this pin
-    ///
-    /// This method is only available, if the pin has been set to dynamic mode.
-    /// See [`Pin::into_dynamic_pin`].
-    /// Unless this condition is met, code trying to call this method will not compile.
-    pub fn get_level(&self) -> Level {
-        match self.is_high() {
-            true => Level::High,
-            false => Level::Low,
-        }
     }
 }
 
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -236,6 +236,29 @@ where
     pub fn inner(&self) -> &P {
         &self.inner
     }
+
+    /// Indicates wether the voltage at the pin is currently HIGH
+    ///
+    /// This can be used when the pin is in any direction:
+    ///
+    /// If it is currently an Output pin, it indicates whether the pin output is set to HIGH
+    /// If it is currently an Input pin, it indicates wether the pin input is HIGH
+    ///
+    /// This method is only available, if two conditions are met:
+    /// - The pin is in the GPIO state.
+    /// - The pin direction is set to input.
+    ///
+    /// See [`Pin::into_input_pin`] and [`into_input`]. Unless both of these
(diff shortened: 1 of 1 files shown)


src/gpio.rs
```diff
@@ -457,6 +475,24 @@ where
...
+    /// This method is only available, if two conditions are met:
+    /// - The pin is in the GPIO state.
+    /// - The pin direction is set to output.
+    ///
+    /// See [`Pin::into_output_pin`] and [`into_output`]. Unless both of these
+    /// conditions are met, code trying to call this method will not compile.
+    ///
+    /// [`Pin::into_output_pin`]: ../pins/struct.Pin.html#method.into_output_pin
+    /// [`into_output`]: #method.into_output
+    pub fn get_set_level(&self) -> Level {
```
hannobraun, line 489: Weird name, but I can't think of a better one off the top of my head. Can be left as-is for now.

src/gpio.rs
```diff
@@ -457,6 +475,24 @@ where
...
+        match self.is_set_high() {
+            true => Level::High,
+            false => Level::Low,
+        }
```
hannobraun, line 493: Suggestion: To reduce code duplication, how about an `impl From<bool> for Level`?

hannobraun (commented):
Thank you, @Lotterleben! Looks good to me, but I left some suggestions. Feel free to address them or not, as time/motivation permits.

CI fails, but this seems unrelated. Looks like we're picking up the latest alpha release of RTIC? This shouldn't happen, as we pin the dependency version. I've made a note to look into it later.

Leaving this open to give @david-sawatzke a chance to take a look. It's currently blocked on the CI failure anyways.

hannobraun:
Can be rebased now.

src/gpio.rs
Lotterleben: Hrrrrm. Sounds enticing at first (declutter! yay!) but from my perspective as an outsider getting to know the codebase, I think this is a potential cause of confusion:
We'd be ascribing semantics to a boolean which aren't intuitively clear if you don't know the backstory:
"Does `level == true` mean it's high because there's current running through it? Or that it's low because as a newcomer, I'm used to "pin low -> LED on"? Oh it's true if `is_high()` is true? Well, I didn't expect *that*"

This could and should of course be documented, but it's still a quirk, and I don't know if it helps with readability even if it removes clutter.

hannobraun: Hmm, yeah, I see what you mean. So how about an inherent method instead? `Level::from_high(bool)`? `Level::from_is_set_high(bool)`? Should reduce duplication all the same, while alleviating your concern.

hannobraun: Or even `Level::read(&GpioPin)`?

src/gpio.rs
Lotterleben: should be resolved by https://github.com/lpc-rs/lpc8xx-hal/pull/311/commits/0ec25f031de19bc2f992d829f0476bc0e8dc44bd, every pin direction uses `is_high` now

src/gpio.rs
Lotterleben: check! implemented `Level::read(&GpioPin)`, but called it `Level:: from_pin()` :)

hannobraun:
Something's wrong. This includes commits that are already in `master`. Did you update your local `master` branch before rebasing?

Lotterleben:
whoops sorry, got tangled in my own branches there. regarding the failing nightly tests... did the error msg format change slightly in nightly?

src/gpio.rs
```diff
@@ -236,6 +236,29 @@ where
...
+    /// conditions are met, code trying to call this method will not compile.
+    ///
+    /// [`Pin::into_input_pin`]: ../pins/struct.Pin.html#method.into_input_pin
+    /// [`into_input`]: #method.into_input
+    pub fn is_high(&self) -> bool {
+        let gpio = unsafe { &*pac::GPIO::ptr() };
+        let registers = Registers::new(gpio);
+
+        is_high(&registers, self.inner())
+    }
```
hannobraun, line 261: This method seems have wandered into the unconditional `impl` block, meaning it is now available regardless of pin direction. That doesn't seem right. Shouldn't it just stay where it is?

hannobraun (changes requested):
Thanks for the changes! One more thing I noticed. Left a comment.

The CI failure with nightly is not critical, I think. The rustc errors must have changed, as you said. We can just ignore it until it makes its way to stable, then we have to update the expected error.

src/gpio.rs
Lotterleben: we can't keep it where it is, because we need to be able to access it from `Level:: from_pin` (or we'd need 3 implementations, for `D: Direction::Input`, `D: Direction::Output` and `D: Direction::Dynamic`.
*However*, I'd argue that this is is not a bug, but actually a feature– maybe you'd want to read the level of an Output pin, to assert that it behaves as you're expecting it to. That'd also enable us to just get rid of the awkwardly-named `get_set_level` … I don't know why I didn't already do that

hannobraun: Ah, okay, I understand. The method is available regardless of state, after all, as there's no way to have an uninitialized `GpioPin`.

So I guess we have to address the following questions/issues:
- Do we want to make the formerly input-only methods available unconditionally? That seems attractive to me, due to the simplicity and because it enables `Level::from_pin`.
- If yes, we should be consistent about it and do this for all input methods (`is_high`, `is_low`, `get_level`).
- If yes, we need to update the documentation, as that's currently wrong.
- If yes, do we want to keep `is_set_high`/`is_set_low`? They would be redundant, but a case could be made that they should be kept, to follow the style of the embedded-hal traits.

@david-sawatzke, what do you think? I'm currently tending towards making `is_high`/`is_low`/`get_level` unconditional and removing `is_set_high`/`is_set_low`, but I'm somewhat undecided.

david-sawatzke: (This is more of a drive-by comment, I haven't thought deeply about this)

I'd tend to keep them both separate, for a couple of reasons:
1. Consistency with the HAL traits
2. If (or when) we add an OD function, they need to be separate
3. Reduce confusion, imagine e.g. having a high load on an output pin and hoping `is_high` could be used to detect it.

I'd also not expose them for output-only pins, so it becomes harder to miss accidentally using a pin in the wrong direction.

Maybe as an alternative then add a `pub(crate)` `fn` unconditionally and implement `Level::form_pin` with that for both input & output (not sure if that's more easily said than done)?

hannobraun: Hmm, yeah, makes sense.

> Maybe as an alternative then add a `pub(crate)` `fn` unconditionally and implement `Level::form_pin` with that for both input & output (not sure if that's more easily said than done)?

Seems like a workable solution. What do you think, @Lotterleben?

Lotterleben: that sounds really good! will look into it tomorrow.

src/gpio.rs
```diff
@@ -236,6 +236,17 @@ where
     pub fn inner(&self) -> &P {
         &self.inner
     }
+
+    /// Indicates wether the voltage at the pin is currently HIGH
+    /// This is not accessible to the user to avoid confusion becauzse `is_high()`
```
hannobraun, line 241:
```suggestion
    /// This is not accessible to the user to avoid confusion because `is_high()`
```

hannobraun (approved):
Looks good, thank you @Lotterleben!

I left one suggestion, but I think I can apply this myself after submitting the review.

PR #312, opened 2021-03-05
Upgrade to latest RTIC alpha

hannobraun:
The build was broken due to the release of the new RTIC alpha version.
We pin a specific RTIC version in Cargo.toml to prevent just that, but
that won't prevent Cargo from happily picking up breaking changes from
transitive dependencies. Since RTIC consists of multiple crates, that is
what happened here.

We could explicitly specify and pin all those transitive dependencies in
Cargo.toml to prevent that from happening, but that would get unwieldy.
I decided that it's probably best to just upgrade to the latest alpha,
ride it out for this alpha cycle, and be more careful about adding alpha
dependencies in the future.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ default-features = false
 
 
 [dev-dependencies]
-cortex-m-rtic = "=0.6.0-alpha.0"
+cortex-m-rtic = "=0.6.0-alpha.1"
 
 [dev-dependencies.panic-rtt-target]
 version  = "0.1.1"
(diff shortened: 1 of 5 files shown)


hannobraun:
#311 is blocked by the problem being addressed here. Once this PR is merged, #311 needs a rebase to pick up this change (cc @Lotterleben).

Lotterleben:
ack

hannobraun:
CI fails on nightly only. It seems changes to compiler output cause the compile tests to fail. Stable is green though, so this should be ready to go.

david-sawatzke (approved):

PR #316, opened 2021-04-20
Update dependencies

hannobraun:
Upgrades to the new versions of the PACs (see https://github.com/lpc-rs/lpc-pac/pull/57). I did some light testing on LPC845 by running some of the examples, and didn't see any problems. Since this is about all I have time for right now, I suggest leaving it at that and letting problems shake out in master (if there are any problems; the SVD files stayed the same, so I don't expect anything major).

Currently blocked on new released of `lpc82x-pac` and `lpc845-pac`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ travis-ci = { repository = "lpc-rs/lpc8xx-hal" }
 
 
 [dependencies]
-cortex-m      = "0.6.4"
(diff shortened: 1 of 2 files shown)


hannobraun:
PACs have been released, this pull request has been updated. Should be good to go now.

david-sawatzke (approved):
Looks Good!

Issue #324, opened 2021-08-10
Build is broken with latest Trybuild version


hannobraun:
See https://github.com/dtolnay/trybuild/issues/122.

hannobraun:
Fixed upstream.

PR #325, opened 2021-08-10
Improve build script output

hannobraun:
This came out of my efforts to track down #324, but seems more broadly useful.


david-sawatzke (approved):
Looks very reasonable

david-sawatzke:
It seems like tests are once again broken due to some weird target issues, but the issue doesn't stem from this pr

hannobraun:
CI breakage is caused by #324. That should be taken care of now, as the latest version of Trybuild has been yanked.

PR #328, opened 2021-09-22
Upgrade to latest PAC versions

hannobraun:
Upgrade to latest versions of the PACs. Tested by running a few of the examples. I don't expect any regressions, since the changes aren't substantial.

Submitted as a draft. I plan to update this to use published versions, once they are out.


hannobraun:
Turns out the new PAC releases were non-breaking point releases, so no need to change anything here. Closing.

PR #329, opened 2021-09-27
Update dependency on cortex-m-rt

hannobraun:
We aren't affected by the breaking changes of the 0.7 release. The PAC
has already been updated to accept both 0.6 and 0.7, so this leaves it
up to the user which version to use.


hannobraun:
The nightly CI build fails because a compiler error message changed. This isn't relevant to this PR. We just need to update the expected error message once the improve error messages makes its way to stable.

